PR #20, opened 2017-05-07
Humidity Sensor

mdaxter:
This adds a humidity sensor and also fixes minor visibility issues where where the current temperature of the temperature sensor was not accessible.  This change also makes the constructor of the Accessory open class public so that can actually be subclassed outside the module.

### DIFF
--- a/Sources/HAP/Accessories/HumiditySensor.swift
+++ b/Sources/HAP/Accessories/HumiditySensor.swift
@@ -12,7 +12,7 @@ public typealias CurrentRelativeHumidity = Double
 
 extension Service {
     open class HumiditySensor: Service {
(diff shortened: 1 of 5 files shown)


Bouke:
Great PR, thank you! How's the library working out for you? How are you using it, if you don't mind me asking?

mdaxter:
It's very nice, I prefer Swift a lot over Go (which I have been using so far for bridging)!  The only thing that still seems to cause some issues is re-connecting, particularly when restarting the HAP server.  This often requires me to completely remove the bridge and re-pair it.

mdaxter:
Further to this, it looks like the re-connection issue is gone now (I removed the persistent storage and re-created it).  I put my SonoffTasmotaHAP project on my github page if you are interested.

PR #24, opened 2017-07-12
Update README.md

moflo:
I've been using Xcode to debug HAP and wanted to share the setup. I'm using your code to debug some embedded C/C++ code: https://github.com/moflo/homekit-accessory-emulator

Thanks!
--nye.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ And then build the project itself:
 
 Run ``swift build`` to compile and ``.build/debug/hap-server`` to run. Modify ``Sources/hap-server/main.swift`` to include your own accessories.
 
+On Mac OS, you can debug using XCode by running the command ``swift package generate-xcodeproj`` and the opening the resulting ``HAP.xcodeproj`` project. Select the ``hap-server`` target to execute.
+
 ## Linux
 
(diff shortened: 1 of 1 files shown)


Bouke:
Ah yes; using Xcode is also how I developed this package. Great to hear that it proves useful for your use case 👍.

moflo:
Yes, thank you. By the way, any tips on debugging? I'm currently just using the Home app and trial & error, sometimes dropping into WireShark to analyze UDP/TCP packets, but always looking for a better way!

Bouke:
I also used `hc/brutella`, written in go. Sometimes when I couldn't explain some failure, I tried reproducing the situation in that implementation. Mostly I found that some part of the crypto setup was incorrect.

The most difficult part was getting Poly1305-Chacha20 correctly. There are two implementations in libsodium: the original by DJB and the IETF variant. However Apple's implementation is somewhat a hybrid between the two.

What helps best is fixating the random inputs (secret keys salt etc) on both ends (if possible) and work your way up from there. After the implementation of a certain feature is working –e.g. pair setup– use random data again and work out the edge cases. Using this approach I found other implementation differences as well, for example how leading zeros are handled in SRP.

PR #27, opened 2017-10-18
Conformance to HAP Specification section 5.7

tgu:
In the process of finding some bugs in my application code I have found some possible changes.
- Value should only and always be present for read permission
- Sometimes the response should be 207 multi-status with an HAP Status Code
- More diverse response messages for write request

### DIFF
--- a/Sources/HAP/Endpoints/characteristics().swift
+++ b/Sources/HAP/Endpoints/characteristics().swift
@@ -18,9 +18,13 @@ func characteristics(device: Device) -> Application {
             }
             
             let meta = queryItems?.first(where: {$0.name == "meta"})?.value == "1"
+            let perms = queryItems?.first(where: {$0.name == "perms"})?.value == "1"
+            let type = queryItems?.first(where: {$0.name == "type"})?.value == "1"
+            let ev = queryItems?.first(where: {$0.name == "ev"})?.value == "1"
 
             let paths = id.components(separatedBy: ",").map { $0.components(separatedBy: ".").flatMap { Int($0) } }
 
+            var multiStatusResponse = false
             var serialized: [[String: Any]] = []
             for path in paths {
                 guard path.count == 2 else {
@@ -30,7 +34,9 @@ func characteristics(device: Device) -> Application {
                     let characteristic = device.accessories.first(where: {$0.aid == path[0]})?.services.flatMap({$0.characteristics.filter({$0.iid == path[1]})}).first
                     else {
                         // @fixme: hc sets status to StatusServiceCommunicationFailure instead
-                        return .notFound
+                        serialized.append(["aid": path[0],"iid": path[1],"status": HAPStatusCodes.resourceDoesNotExist.rawValue])
+                        multiStatusResponse = true
+                        break
                 }
                 
                 var body = ["aid": path[0],"iid": path[1],"value": characteristic.getValue() ?? NSNull()]
@@ -51,13 +57,30 @@ func characteristics(device: Device) -> Application {
                         body["maxLen"] = maxLen.jsonValueType
                     }
                 }
+                if perms {
+                    body["perms"] = characteristic.permissions.map { $0.rawValue }
+                }
+                if type {
+                    body["type"] = characteristic.type.rawValue
+                }
+                if ev {
+                    body["ev"] = characteristic.permissions.contains(.events)
+                }
                 
                 serialized.append(body)
+            }
+            
+            if multiStatusResponse {
+                for (index,element) in serialized.enumerated() {
+                    if element["status"] == nil {
+                        serialized[index]["status"] = HAPStatusCodes.success.rawValue
+                    }
+                }
             }
 
             do {
                 let json = try JSONSerialization.data(withJSONObject: ["characteristics": serialized], options: [])
-                return Response(data: json, mimeType: "application/hap+json")
+                return Response(status: multiStatusResponse ? .multiStatus : .ok, data: json, mimeType: "application/hap+json")
             } catch {
                 logger.error("Could not serialize object", error: error)
                 return .internalServerError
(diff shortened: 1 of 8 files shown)


Sources/HAP/HTTP/Message.swift
```diff
@@ -40,11 +42,14 @@ class Response {
...
             case .accepted: return "202 Accepted"
             case .noContent: return "204 No Content"
+            case .multiStatus: return "207 Multi-Status"
             case .movedPermanently: return "301 Moved Permanently"
             case .badRequest: return "400 Bad Request"
             case .unauthorized: return "401 Unauthorized"
             case .forbidden: return "403 Forbidden"
             case .notFound: return "404 Not Found"
+            case .methodNotAllowed: return "405 Method Not Allowed"
+            case .invalidParameters: return "422 Unprocessable Entity"
```
Bouke: Should we call it `unprocessableEntity` instead of `invalidParameters`, similar to the status' description?

Sources/HAP/Endpoints/characteristics().swift
```diff
@@ -89,15 +146,32 @@ func characteristics(device: Device) -> Application {
...
                         logger.info("Removed listener for \(characteristic)")
                     }
+                    serialized.append(["aid": aid,"iid": iid,"status": HAPStatusCodes.success.rawValue])
                 }
             }
+            
+            if multiStatusResponse {
+                do {
+                    let json = try JSONSerialization.data(withJSONObject: ["characteristics": serialized], options: [])
+                    return Response(status: serialized.count == 1 ? .badRequest : .multiStatus, data: json, mimeType: "application/hap+json")
```
Bouke: Instead of setting `multiStatusResponse`, can't we keep an array of responses? And when it's time to create the response, see if there's a common denominator which we can return and `.multiStatus` otherwise? Is there a specific reason you introduced this state flag?

Sources/HAP/Endpoints/characteristics().swift
```diff
@@ -65,11 +113,18 @@ func characteristics(device: Device) -> Application {
...
                 {
-                    return .notFound
+                    return .invalidParameters
                 }
 
                 // set new value
                 if let value = item["value"] {
+                    guard characteristic.permissions.contains(.write) else {
+                        logger.info("\(characteristic) has no write permission")
+                        serialized.append(["aid": aid,"iid": iid,"status": HAPStatusCodes.readOnly.rawValue])
```
Bouke: Please put a space after the comma for readability.

Sources/HAP/Endpoints/characteristics().swift
```diff
@@ -25,18 +34,53 @@ func characteristics(device: Device) -> Application {
...
+                    body["ev"] = characteristic.permissions.contains(.events)
+                }
+                
+                serialized.append(body)
+            }
+            
+            if multiStatusResponse {
+                for (index,element) in serialized.enumerated() {
+                    if element["status"] == nil {
+                        serialized[index]["status"] = HAPStatusCodes.success.rawValue
```
Bouke: I'm not familiar with section 5.7 of the HAP spec yet. Do we only need to set the `status` when we're doing a `multiStatus` response?

tgu: The specs are not clear on this. The examples, hovewer, suggest that the status should be omitted where there are no errors.

Sources/HAP/Endpoints/characteristics().swift
tgu: Fixed

Sources/HAP/Endpoints/characteristics().swift
tgu: See your point, flags are difficult to maintain.

Sources/HAP/HTTP/Message.swift
tgu: Makes sense!

tgu:
I think I can provide some tests.

Tests/HAPTests/EndpointTests.swift
```diff
@@ -202,9 +204,407 @@ class EndpointTests: XCTestCase {
...
+            
+            guard let therm = characteristics.first(where: { $0["aid"] as? Int == thermostat.aid }) else {
+                return XCTFail("Could not get therm aid")
+            }
+            
+            guard let lampa = characteristics.first(where: { $0["aid"] as? Int == lamp.aid }) else {
+                return XCTFail("Could not get lampa aid")
+            }
+            
+            guard let lightVal = light["value"] as? Double else {
```
Bouke: On Linux, decoding from JSON is a bit more involved and that's why the test is probably failing here. When there's no decimal separator in the serialised value, SwiftFoundation decodes it as an Int, instead of a Double. See also [SR-4599](https://bugs.swift.org/browse/SR-4599). I've resolved it in HAP by also supporting initialisation of Int in 25a7afd9e8948ff47db285dfe1b3d5249f563f29, which is used at [Characteristic.swift#L71](https://github.com/Bouke/HAP/blob/ccd4e263adce8993f39394bed7b3a026bc3c133c/Sources/HAP/Base/Characteristic.swift#L71).

Bouke: Seeing that `Double(value: Any)` is public, you should be able to rewrite this line (and the others below) as:

    guard let lightVal = Double(value: light["value"]) else {

tgu: Thanks! I had difficulties to understand the reason for the error, uploaded 95c58ef43d1edf6ec17f6f770bed2ee9d8ce39cc that hopefully will resolve the problem. I have currently no Linux installation so I can't test the code.

Bouke:
Great work! I've merged your PR in d7ce5f498fac70792af5a8b1468d49281b26b6a2.

Issue #32, opened 2017-10-30
Persist device config


Bouke:
See also the specs for `c#` and other mDNS fields. Also; should we serialize the instance IDs so we can re-assign them? Should we support adding/removing accessories dynamically?

gbrooker:
Addressed in pull request #40

PR #33, opened 2017-10-31
iid of info should always be 1

tgu:
HAP Spec 2.6.1.2 Service and Characteristic Instance IDs

> The Accessory Information service must have a service instance ID of 1.

This seems to be critical. The Home app refuses to connect otherwise

### DIFF
--- a/Sources/HAP/Server/Device.swift
+++ b/Sources/HAP/Server/Device.swift
@@ -81,7 +81,7 @@ public class Device {
             accessory.device = self
             accessory.aid = idGenerator.next()!
             for service in accessory.services {
-                service.iid = idGenerator.next()!
+                service.iid = service is Service.Info ? 1 : idGenerator.next()!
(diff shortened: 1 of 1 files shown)


Bouke:
Also from 2.6.1.2:
> After a firmware update services and characteristics types that remain unchanged must retain their previous instance ids, newly added services and characteristics must not reuse instance ids from services and characteristics that were removed in the firmware update.
In #32 I'm thinking of how to implement this requirement. We could re-use the database to create a persistent mapping of instance IDs, and assigning new IDs when certain parameters change.

Bouke:
I'll revisit the assignment of instance IDs as well; my change of yesterday uses a global pool, but that's not required by the specs.

Bouke:
I've committed a fix in 3e614755cf28378f41164ead433a879e70bf1759, along with a test.

PR #36, opened 2017-12-28
Add Sensor accessories

gbrooker:
Add implementations of three Sensor accessories, and a new method for Accessory to permit subclasses to add additional services


Bouke:
I'd like to merge the added sensor accessories, however the API change in `Accessory` I'm not so sure about. By subclassing you could already override the initializer I think, and then inject the services you'd like to add. The `Accessory` would ideally be a struct, as it should not be mutable after initialization.

gbrooker:
Hi,

Without this change, it is not possible to override an implemented accessory, a Switch or Door for example, and add other services, which is permitted under the HAP spec. The alternative would be to have an optionalServices parameter to the init() method of every Accessory implementation, which would need changes to every accessory class.

Thanks very much for your HAP implementation, I’m having a lot of fun experimenting with it. Happy to take a look at anything on your todo list if you would like me to. Otherwise I plan to add a couple of concrete implementations of a few devices I currently control though Homebridge.

Cheers,
Guy

Bouke:
Ah I see what you mean. However instead of instantiating a derived `Accessory`, you can instantiate the base class `Accessory`. This gives you additional freedom to specify any additional service you'd like:

```swift
// this
let accessory = ContactSensor(info: .info(...))

// becomes:
let contactSensor = Service.ContactSensor()
let accessory = Accessory(info: .info(...), type: .sensor, services: [contactSensor, ...])
```

On a side nodeThere's also the concept of *optional services* in the HAP specification, which I'd like to have some support for. These services would be instantiated optionally by the derived `Accessory` classes.

I'd be happy to merge other items on the todo list 👍.

Sources/HAP/Server/Server.swift
```diff
@@ -91,6 +91,16 @@ public class Server: NSObject, NetServiceDelegate {
                         break
                     }
 
+                    // Fix to allow Bridges with spaces in name
+                    // HomeKit POSTs contain a hostname with \\032 replacing the space
+                    // which causes Kitura httpParser to baulk on the resulting URL.
+                    // Replacing '\\032' with '-' in the hostname allows Kitura to create a valid URL,
+                    // and the value is not used elsewhere.
```
Bouke: Great to see a fix for this. However I'd prefer to have separate fixes in separate PRs, so we can review/discuss them separately.

Sources/HAP/Server/Server.swift
```diff
@@ -91,6 +91,16 @@ public class Server: NSObject, NetServiceDelegate {
...
+                    //
+                    if let hostname = request.headers["Host"]?[0], (hostname.contains("\\032")) {
```
Bouke: `"\\032"` could be written as `" "` which is clearer, I think.

Sources/HAP/Server/Server.swift
```diff
@@ -91,6 +91,16 @@ public class Server: NSObject, NetServiceDelegate {
...
+                        request.headers["Host"]![0] = hostname.replacingOccurrences(of:"\\032", with:"-")
```
Bouke: I don't think we can easily add a test for this? We should –probably separately from this bugfix– add some test infrastructure to test the basic http server edge cases.

Bouke:
> You are right it would be possible to reimplement an accessory to add new services, but wouldn’t it be better to have a design where more sophisticated accessory implementations subclass the basic accessory implementations provided ?

> If you don’t like the subclass driven approach using addServices(), perhaps it could be refactored as a function subclassServices() -> [Service] called by Accessory.init() whose basic implementation would return [], and subclasses would override the function to return the array from the superclass implementation with additional services appended.

I feel like the current API is sufficient for either use-case; you can either:

* instantiate a generic `Accessory` and specify any service you'd like to add, to cater for ad-hoc service composition; or
* subclass `Accessory` (or one of its subclasses) and inject any additional services by overriding `init`.

> and will look at dynamically adding accessories to the bridge, pre and post pairing.
I'm not sure if this is even allowed in the HAP specification?

> I’ll also look at preserving id’s for accessories between launches, unless you’ve already got one or more of those covered yourself.
I was thinking of using `Codable` to serialize accessories and use the serial number as the stable identifier to compare against. So changing the serial number would also result in changing the identifiers.

gbrooker:
> On 8 Jan 2018, at 01:37, Bouke Haarsma <notifications@github.com> wrote:
> 
> You are right it would be possible to reimplement an accessory to add new services, but wouldn’t it be better to have a design where more sophisticated accessory implementations subclass the basic accessory implementations provided ?
> 
> If you don’t like the subclass driven approach using addServices(), perhaps it could be refactored as a function subclassServices() -> [Service] called by Accessory.init() whose basic implementation would return [], and subclasses would override the function to return the array from the superclass implementation with additional services appended.
> 
> I feel like the current API is sufficient for either use-case; you can either:
> 
> instantiate a generic Accessory and specify any service you'd like to add, to cater for ad-hoc service composition; or
> subclass Accessory (or one of its subclasses) and inject any additional services by overriding init.
It is indeed possible to do the first bullet, but the current implementation doesn’t permit additional Services to be added by a subclass of the provided accessories, such as Door, or Outlet, because their init()’s do not include a services parameter, and there is currently no other way for a subclass to add services other than through Accessory.init()
> and will look at dynamically adding accessories to the bridge, pre and post pairing.
> 
> I'm not sure if this is even allowed in the HAP specification?

It is indeed, see Clause 5.6.2. The bridge’s configuration number c# needs to be updated. homebridge already supports this.
> I’ll also look at preserving id’s for accessories between launches, unless you’ve already got one or more of those covered yourself.
> 
> I was thinking of using Codable to serialize accessories and use the serial number as the stable identifier to compare against. So changing the serial number would also result in changing the identifiers.

I see, so you would plan to replace the JSONSerializable protocol/methods ?

gbrooker:
will do

gbrooker:
I've moved the bridge name fix and the addServices changes to two separate branches as requested.

Bouke:
Thanks!

PR #38, opened 2018-01-08
Fix to allow Bridges with spaces in name

gbrooker:


Sources/HAP/Server/Server.swift
```diff
@@ -91,6 +91,16 @@ public class Server: NSObject, NetServiceDelegate {
                         break
                     }
 
+                    // Fix to allow Bridges with spaces in name
+                    // HomeKit POSTs contain a hostname with \\032 replacing the space
+                    // which causes Kitura httpParser to baulk on the resulting URL.
+                    // Replacing '\\032' with '-' in the hostname allows Kitura to create a valid URL,
+                    // and the value is not used elsewhere.
+                    //
+                    if let hostname = request.headers["Host"]?[0], (hostname.contains("\\032")) {
```
Bouke: `"\\032"` could be written as `" "` which is clearer, I think.

gbrooker: Actually "\\032" is not equivalent to " ". For a Bridge advertising itself in mDNS as '"Bridge One"' iOS is really sending

    POST /pair-setup
    Host: Bridge\032One._hap._tcp.local

Most HAP servers would probably ignore the Host header, but Kitura insists on building a URL as it processes the reply, which causes the /pair-setup request to be rejected. Another solution would be to drop the dependancy on Kitura, and implement the http request/reply directly.

Bouke:
Good to know, thanks!

PR #39, opened 2018-01-08
Allow subclasses of predefined accessories to add Services

gbrooker:
Add an addServices() method to permit subclasses of predefined accessories to add additional services

### DIFF
--- a/Sources/HAP/Base/Accessory.swift
+++ b/Sources/HAP/Base/Accessory.swift
@@ -26,19 +26,26 @@ open class Accessory {
     internal var aid: InstanceID = 0
     public let type: AccessoryType
     public let info: Service.Info
-    internal let services: [Service]
-
+    internal var services: [Service]
+    
+    // TODO: SwiftFoundation in Swift 4.0 cannot encode/decode UInt64,
+    // which is the data-type we wanted to use here. We can change it back
+    // to UInt64 once the following commit has made it into a release:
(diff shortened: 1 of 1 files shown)


Bouke:
> It is indeed possible to do the first bullet, but the current implementation doesn’t permit additional Services to be added by a subclass of the provided accessories, such as Door, or Outlet, because their init()’s do not include a services parameter, and there is currently no other way for a subclass to add services other than through Accessory.init()

I've pushed commit fe65f3af4088898bc42b26566a7e92c1dfb62e62, which now allows you to do the following:

```
class MyThermostat: HAP.Accessory.Thermostat {
    let airQualitySensor = HAP.Service.AirQualitySensor()

    public init(info: Service.Info) {
        super.init(info: info, additionalServices: [airQualitySensor])
    }
}
```

There isn't support for dynamically adding/removing accessories and services at the moment. This would need additional work in various places; updating the device's "configuration version", removing notification listeners and the like. And meanwhile the API should be easy to use, allow for flexibility but enforce correctness and adherence to the specification.

gbrooker:
Thanks, that works too. We will also need to add the additionalServices  parameter to the Accessory classes I added for a bunch of sensors.

Bouke:
See b1233482518cc98d4fc171ffdedf53e19a6618f6.

gbrooker:
No need for this change now

PR #40, opened 2018-01-17
AID Persistence, Add/Remove Accessories, Reachability

gbrooker:
Added ability to add and remove accessories from the Bridge (while paired or not), and persist the Accessory IDs over restarts of the host application. The configuration number (in mDNS TXT record), the last allocated AID, and the AID's of recognised accessories are written to a configuration key in the Device Storage.

The TXT record broadcast over mDNS is updated when the configuation changes, or the device is paired/unpaired.

Added a reachable flag to Accessory, for subclasses to indicate that the bridge has lost contact with a device. HomeKit will recognise the device is unreachable the next time it tries to read or write to the device. For more rapid updates, add a BridgingState Service to the Accessory. The subclass only has to set the reachable property, if the device has a Bridging State Service, it is updated too. 

Pairing/unpairing is now performed through functions on the Device instance, so the status can be monitored.


Sources/HAP/Base/Accessory.swift
```diff
@@ -30,6 +35,29 @@ open class Accessory {
...
     public let info: Service.Info
     internal let services: [Service]
+    
+    // An accessory implementation can set this flag to false if a device becomes unreachable for a period
+    // If the accessory has a BridgingState Service, then the reachable property of that Service is set
+    //
+    open var reachable = true {
+        didSet {
+            if self.reachable != oldValue,
+               let bridgingStateService = services.first(where: { $0.type == .bridgingState }) as! Service.BridgingState? {
```
Bouke: The `as!` might fail, as having a `type == .bridgingState` does not require an instance of `BridgingState`. So best to make the code all generic, and query the generic accessory for the expected characteristic.

Sources/HAP/Base/Accessory.swift
```diff
@@ -30,6 +35,29 @@ open class Accessory {
...
+                bridgingStateService.reachable.value = self.reachable
+            }
+        }
+    }
+    
+    // An accessory must provide a text identifier, which is guranteed to be unique.
```
Bouke: We can rely on the serial number, and thus require the serial number to be unique (as it should be).

Sources/HAP/Server/Device.swift
```diff
@@ -47,9 +47,81 @@ public class Device {
...
+    //    uniquely identify HAP accessory objects within an HAP accessory server, or uniquely identify
+    //    services, and characteristics within an HAP accessory object. The instance ID for each object
+    //    must be unique for the lifetime of the server/ client pairing.
+    
+    struct Configuration : Codable {
+        var number: UInt32 = 0
+        var lastAID: InstanceID = 1
+        var aidForAccessoryUniqueId = [String : InstanceID]()
+        
+        // The next aid - should be checked against existing devices to ensure it is unique
```
Bouke: I think this better belongs in a function, or refactor it to have a signature like `aidGenerator: () -> InstanceID`

Sources/HAP/Server/Device.swift
```diff
@@ -47,9 +47,81 @@ public class Device {
...
+    //    then no other Accessory object can have an instance ID of "1" within the Accessory Server.
+    //
+    //    Instance IDs are numbers with a range of [1, 18446744073709551615]. These numbers are used to
+    //    uniquely identify HAP accessory objects within an HAP accessory server, or uniquely identify
+    //    services, and characteristics within an HAP accessory object. The instance ID for each object
+    //    must be unique for the lifetime of the server/ client pairing.
+    
+    struct Configuration : Codable {
+        var number: UInt32 = 0
+        var lastAID: InstanceID = 1
```
Bouke: What should be the scope of this variable? We probably don't want this to be changed, outside of the generator and when initializing?

Sources/HAP/Server/Device.swift
```diff
@@ -47,9 +47,81 @@ public class Device {
...
+                if lastAID < 2 {
+                    lastAID = 2
+                }
+                return lastAID
+            }
+        }
+        
+        func writeTo(_ storage: Storage) {
+            do {
+                let encoder = PropertyListEncoder()
```
Bouke: We could use JSON instead, from the Travis build failure it looks like PropertyListEncoder is not available on Linux?

Sources/HAP/Base/Accessory.swift
gbrooker: Very good point. I'll address this. BTW, why does the setValue() func in Characteristic protocol have a fromConnection parameter ? It doesn't seem to be used.

Sources/HAP/Base/Accessory.swift
gbrooker: Can do, but in that case we should change the init for the Service.Info class to remove the default parameter for serial number 

    serialNumber: String = "undefined"

and perhaps add a check in Device to ensure there are no duplicate serial numbers.

Sources/HAP/Server/Device.swift
gbrooker: This structure and all properties and func can be internal, and lastAID can be private

Sources/HAP/Server/Device.swift
gbrooker: Could you elaborate on what you envisage here ?

Sources/HAP/Server/Device.swift
gbrooker: OK, should be straightforward, I'll take a look at that

Sources/HAP/Server/Device.swift
Bouke: Instead of having a mutating getter on a variable, it would be better to have a generator function instead. It would be easier to understand what the code does, and communicate to the internal user that it's a generator.

Sources/HAP/Base/Accessory.swift
Bouke: Yeah we would need the user to explicitly specify a serial number.

Sources/HAP/Server/Server.swift
```diff
@@ -167,15 +167,36 @@ public class Server: NSObject, NetServiceDelegate {
...
 
         service = NetService(domain: "local.", type: "_hap._tcp.", name: device.name, port: socket.listeningPort)
+        
+        Server.publishDiscoveryRecordOf(device, to: service)
+
+        super.init()
+        service.delegate = self
+        
+        device.onConfigurationChange.append( { (device) in
+            Server.publishDiscoveryRecordOf(device, to: self.service)
```
Bouke: This would introduce a retain cycle from `device` to `self.service` (and maybe to the device itself?). Also, this callback will outlive the instance of Server, meaning that even after stopping the server, the callback would continue to publish updates of the device.

There are probably more instances of this problem to be found in the existing code, but we should aim for reducing those.

Sources/HAP/Server/Device.swift
```diff
@@ -144,15 +230,107 @@ public class Device {
...
+        
+    }
+    
+    public func removeAccessories(_ unwantedAccessories: [Accessory]) {
+        
+        var configurationUpdated = false
+        
+        for accessory in unwantedAccessories {
+            // Ensure the initial accessory is not removed, and that the accessory is in the list
+            if accessory.aid != 1 {
```
Bouke: Why shouldn't we remove the initial accessory? Also; we shouldn't silently ignore that, but trap as it is invalid API usage. Thus use something like `preconditionFailure()`.

Sources/HAP/Server/Device.swift
```diff
@@ -144,15 +230,107 @@ public class Device {
...
+                if let i = accessories.index(where: { $0 === accessory}) {
```
Bouke: Would be clearer if you wrote `guard let index = ...`. And again, we should trap on invalid API usage (`preconditionFailure()`).

Sources/HAP/Server/Device.swift
```diff
@@ -144,15 +230,107 @@ public class Device {
...
+                    configuration.aidForAccessoryUniqueId.removeValue(forKey: dsuid)
+                    
+                    configurationUpdated = true
+                }
+            }
+        }
+        
+        // write configuration data to persist updated aid's
+        
+        if configurationUpdated {
```
Bouke: Calling this method should always update the configuration, as we eliminated most cases where we didn't;

* `unwantedAccessories` has invalid entries -> trap
* `unwantedAccessories` could be empty -> we could check beforehand, or maybe even also trap?

Sources/HAP/Base/Accessory.swift
gbrooker: I have renamed the property in Accessory to 'uniqueSerialNumber', and by default look for it from the info record. An implementation can either set the info record or override the property. 

In Device.addAccessories, each Accessory is checked against all Accessories previously added to ensure the new device has a unique serial number. If not, it is not added and a message sent to logger. 

I prefer not to use a precondition, fatal error or exception because implementations may be adding devices by network discovery, and for whatever reason may end up with a two devices with the same serial number, so in such a case I would not want the app to crash.

Sources/HAP/Server/Device.swift
gbrooker: OK I've added an aidGenerator

Sources/HAP/Server/Device.swift
gbrooker: According to the spec "2.5.3.3 Primary HAP Accessory Object", the first accessory exposed by a Bridge has to be the bridge itself, so cannot be removed. The called should create a new Device instead. I've modified the func to use precondition()

Sources/HAP/Server/Device.swift
gbrooker: OK, modified the iterator accordingly

Sources/HAP/Server/Device.swift
gbrooker: OK

Sources/HAP/Server/Server.swift
gbrooker: I've made self weak, and test for nil, and changed the name of device to 'changedDevice' to make it clear it is not implicitly self.device. I hope that fixes the retain cycle

gbrooker:
Thanks for all the thoughtful comments. I think I've addressed all the issues you have raised so far. Yes I like lots of white space, must be Jony Ive's influence ;)

Let me know if you spot anything else.

Sources/HAP/Base/Accessory.swift
```diff
@@ -1,5 +1,10 @@
 import Foundation
 
+// TODO: SwiftFoundation in Swift 4.0 cannot encode/decode UInt64,
+// which is the data-type we wanted to use here. We can change it back
+// to UInt64 once the following commit has made it into a release:
+// https://github.com/apple/swift-corelibs-foundation/commit/64b67c91479390776c43a96bd31e4e85f106d5e1
+
```
Bouke: No need for empty line

Sources/HAP/Base/Accessory.swift
```diff
@@ -30,6 +35,34 @@ open class Accessory {
     public let type: AccessoryType
     public let info: Service.Info
     internal let services: [Service]
+    
+    // An accessory implementation can set this flag to false if a device becomes unreachable for a period
```
Bouke: Please wrap comments (throughout) at 80 characters and end sentences with a period.

Sources/HAP/Base/Accessory.swift
```diff
@@ -30,6 +35,34 @@ open class Accessory {
...
+    // If the accessory has a BridgingState Service, then the reachable property of that Service is set
+    //
```
Bouke: No need for empty comment line

Sources/HAP/Base/Accessory.swift
```diff
@@ -30,6 +35,34 @@ open class Accessory {
...
+    
+    // An accessory must provide a text identifier, which is guranteed to be unique.
+    // The implementation must could provide the actual serial number of a device,
+    // or a MAC address or some other identifier which is not used on any other accessory
+    //
+    // Device will check to ensure the serial numbers of all accessories added to
+    // a bridge are are unique
+    //
+    // This is used for persistance of HomeKit AID's
+
```
Bouke: No need for empty line

Sources/HAP/Base/Accessory.swift
Bouke: We should use a trap on invalid API usage. Logging an error is not visible enough to the consumer. Also printing a log message does not trigger some programming logic on the consumer which can be used to resolve this error.

I understand it could possible that duplicate serial numbers might occur. However the consumer should ensure that they are unique, thus could query the `Device` to see if there's already an accessory with that serial number. If there's already an accessory with that serial number, the consumer should decide what to do; either ensure the serial number is unique, or not add the accessory at all.

Sources/HAP/Server/Device.swift
```diff
@@ -47,9 +47,87 @@ public class Device {
...
+        mutating func next() -> InstanceID? {
+            lastAID = lastAID &+ 1 // Add one and overflow if reach max InstanceID
+            if lastAID < 2 {
+                lastAID = 2
+            }
+            return lastAID
+        }
+    }
+    
+    internal struct Configuration : Codable {
```
Bouke: No space before the colon.

Sources/HAP/Base/Accessory.swift
gbrooker: OK, I have revised Device.addAccessories to return an array of Accessories that were successfully added, and also provided a function canAddAccessory() which can be used to check beforehand. I've also added a couple of preconditions to addAccessory, to check the device is a bridge and that the upper limit of 99 accessories (100 including the bridge) will not be exceeded.

gbrooker:
I think I have addressed all your comments now, including the blank lines and comments within 80 characters. 

However I really struggled with git & Xcode source control tonight to commit these last changes due to merging conflicts in Device, Server and characteristics(). I am afraid that some of the improvements you have added in characteristics().swift has conflicted with some of the changes I made. I think the latest version on that file on GitHub has lost a couple of my fixes. I need to double check this tomorrow.

The only comment I didn't address here was the Pairing struct. I think it is a good idea, but I would prefer to work on that in another branch, once these commits are complete. I have a couple of other additions to propose which related to those instance variables. Let's get this dynamic accessory complete and then I'll offer you a new pull request.

gbrooker:
Actually it seems the opposite has happened, my changes appear to have overwritten your changes to characteristic().swift, even though I manually did the opposite in the resolve conflict process. Git really gives me a headache.... If there is something I need to do on my end, please let me know.

Sources/HAP/Base/Accessory.swift
Bouke: The parameter is used in `characteristics().swift` and makes sure that the originating connection doesn't get updated of a change it made itself.

Sources/HAP/Base/Accessory.swift
```diff
@@ -31,6 +35,37 @@ open class Accessory: JSONSerializable {
...
+    // An accessory must provide a text identifier, which is guranteed to be
+    // unique. The implementation could provide the actual serial number of a
+    // device, or a MAC address or some other identifier which is not used on
+    // any other accessory.
+    //
+    // Device will check to ensure the serial numbers of all accessories added
+    // to a bridge are are unique.
+    //
+    // This is used for persistance of HomeKit AID's.
+    open var uniqueSerialNumber: String {
```
Bouke: We should call this `serialNumber`, dropping the `unique` prefix. Naming something a serial number is already implying uniqueness.

Sources/HAP/Base/Accessory.swift
```diff
@@ -31,6 +35,37 @@ open class Accessory: JSONSerializable {
...
+        let serialNumber = info.serialNumber.value
+        precondition(serialNumber != nil)
```
Bouke: Don't think we really need the precondition here -- the serial number is a mandatory field. However the API doesn't reflect this.

Sources/HAP/Server/Device.swift
```diff
@@ -48,9 +50,92 @@ public class Device {
     public let setupCode: String
     let storage: Storage
     let pairings: Pairings
-    public let accessories: [Accessory]
+    public var accessories: [Accessory]
     internal var characteristicEventListeners: [Box<Characteristic>: WeakObjectSet<Server.Connection>]
     public var onIdentify: [(Accessory?) -> Void] = []
+    internal var configuration: Configuration
+    public var onConfigurationChange: [(Device) -> Void] = []
+    public let isBridge: Bool
```
Bouke: We could change this into a computed property:

```swift
public let isBridge: Bool {
    return accessories[0].type == .bridge
}
```

Sources/HAP/Server/Device.swift
```diff
@@ -48,9 +50,92 @@ public class Device {
...
+        internal var aidForAccessorySerialNumber = [String: InstanceID]()
+        private var aidGenerator = AIDGenerator()
+
+        // The next aid - should be checked against existing devices to ensure it is unique
+        internal mutating func nextAID() -> InstanceID {
+            return aidGenerator.next()!
+        }
+
+        // Write the configuration record to storage
+        internal func writeTo(_ storage: Storage) {
```
Bouke: I don't like wrapping this inside `Storage`, and having reading/writing in separate objects. However I'm fine with this for now, as it's likely to change and we can move all storage into `Configuration` and drop `Storage`.

Sources/HAP/Server/Device.swift
```diff
@@ -145,15 +200,141 @@ public class Device {
...
+               !isUniqueAID(accessory.aid, ignoring: accessory) {
+                // swiftlint:disable:next line_length
+                logger.info("Accessory \(accessory.info.name.value ?? "unknown") has a duplicate accessory ID \(accessory.aid), replacing with fresh IID")
+                accessory.aid = 0
+            }
+            if accessory.aid == 0 {
+                // Obtain a new aid, which has not already been used
+                repeat {
+                    accessory.aid = configuration.nextAID()
+                } while (!isUniqueAID(accessory.aid, ignoring: accessory))
```
Bouke: Why do we need to do this? As I understand it, the configuration's last assigned AID is restored on startup. This AID will only increase, so each AID that was handed out, should then never be reused. The spec also says that the AID should be left unchanged for the remainder of the pairing. So my thinking is that we either;

* restore the AID from the configuration, if possible
* assign a new AID from the generator -- no need to check for duplicates, I think?

Sources/HAP/Base/Accessory.swift
```diff
@@ -31,6 +35,37 @@ open class Accessory: JSONSerializable {
     public let info: Service.Info
     internal let services: [Service]
 
+    // An accessory implementation can set this flag to false if a device
+    // becomes unreachable for a period. If the accessory has a
+    // BridgingState Service, then the reachable property of that Service is set.
+    open var reachable = true {
```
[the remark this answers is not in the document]
gbrooker: Yes definitely, an accessory can be reachable or not, even if it doesn't expose a BridgingState characteristic. The reachability is communicated to the HAP controller via the status code to a get or put request (Spec 5.7.1.4)

Sources/HAP/Base/Accessory.swift
gbrooker: Can do

Sources/HAP/Base/Accessory.swift
gbrooker: Perhaps you would prefer to put a condition on Info in that case, and remove the default "undefined" value.

Sources/HAP/Server/Device.swift
gbrooker: I purposely didn't make it a computed property for efficiency reasons. Once the test has been made at initialisation, there is no need to recomputed it, as the instance can never change type.

Sources/HAP/Server/Device.swift
gbrooker: Yes I agree, let's do that at some stage.

Sources/HAP/Server/Device.swift
gbrooker: The spec requires AID's to be unique. In order to avoid a HomeKit database issue, I felt it wise to aggressively ensure that. Indeed the logic is designed that an AID should be unique, however as it is relying on values stored in a configuration, which could have been corrupted or modified. Therefore I felt it was safer to ensure the uniqueness for each accessory added, even one which has been seen before.

gbrooker:
Finally got my MacBook back tonight with a brand new battery. I've fixed the issues raised by swiftlint and I corrected the error that caused the unit tests to fail (precondition in Device.addAccessories needed to allow for case of a single Accessory device). Hope we are now close to a merge to master.

Sources/HAP/Server/Device.swift
```diff
@@ -145,15 +200,133 @@ public class Device {
...
+    ///
+    /// It is an error to try and add accessories with duplicate serial numbers.
+    /// It is an error to try and add accessories to a non-bridge device.
+    /// It is an error to try and increase the number of accessories above 99.
+    public func addAccessories(_ newAccessories: [Accessory]) {
+        let totalNumberOfAccessories = accessories.count + newAccessories.count
+        precondition(
+            (isBridge && totalNumberOfAccessories <= 100) ||
+            (!isBridge && totalNumberOfAccessories == 1),
+                     "A maximum of 99 accessories can be added to a bridge")
```
Bouke: Should we have the following two preconditions instead -- we don't want to add accessories to a non-bridge device:
```swift
precondition(isBridge)
precondition(totalNumberOfAccessories <= 100)
```

Sources/HAP/Server/Device.swift
```diff
@@ -47,9 +47,87 @@ public class Device {
     public let setupCode: String
     let storage: Storage
     let pairings: Pairings
```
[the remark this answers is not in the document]
Bouke: Please change this.

Sources/HAP/Base/Accessory.swift
Bouke: That's a good idea. Let's make the serial number a required argument, so remove the default argument value.

Sources/HAP/Base/Accessory.swift
Bouke: Please change this.

Sources/HAP/Server/Device.swift
Bouke: Ok

Sources/HAP/Server/Device.swift
Bouke: Ok

Sources/HAP/Server/Device.swift
```diff
@@ -174,6 +380,32 @@ public class Device {
...
+    public func removePairing(_ pairingKey: Data) {
+        pairings[pairingKey] = nil
+        
+        if !isPaired {
+            // Update the Bonjour TXT record
+            notifyConfigurationChange()
+        }
+    }
+
+    // Characteristing listeners
```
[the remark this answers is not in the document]
Bouke: Please change this.

Sources/HAP/Server/Device.swift
gbrooker: In Device.init(), all accessories, including the primary accessory are added with this function, therefore it needs to handle the case of a single non-bridge device being added at initialization. This precondition was the reason the unit tests failed.

gbrooker:
OK, have resolved all of those as you requested, except for the precondition in addAccessories(), due to the explanation given above. There were a lot of changes to the test files, as every device/accessory initialiser needed a serial number. I have used a few digits to give each one something unique, you may prefer to use something different such as mac style addresses to serve as better examples.

As discussed, the next pull request I plan would be 

- to replace Storage completely with a single JSON configuration file
- to put paring related properties into a struct
- and I will also add setupCode functions, to permit auto-pairing with a QR code, I have that working on my local implementation, and absolutely love it !

Let me know if there is anything else you think needs urgently addressing.

Bouke:
Thanks! I've merged the PR with a few minor changes (see https://github.com/Bouke/HAP/pull/40/commits/20832f40b4d1454176907edaaa8e02f74cb56107).

Bouke:
See 87204904a23d0ba93b4158b237581e437fcfc061 where I've combined all device settings into the configuration. 

It would be cool to have QR pairing available. I've looked into this before, but I couldn't find a nice way to display QR codes in the terminal. What solution are you thinking about?

Re the storage; we currently use `MemoryStorage` in the tests. It would good for tests to use some memory-backed configuration storage instead of file-based. That allows to easily verify the actual configuration from within the tests, without the need to read the files. I haven't though about how that case should be handled.

Bouke:
FYI I've simplified storage in b55c934ebad0ecb26f71550762f6bf66703730b3.

gbrooker:
Thanks for the merge, and the cleanup. Nice simplification of Storage. I note that the--recreate option in your simple main func is removing the configuration entirely, not just removing the pairings, so the AID and IID's are all lost.

QR code pairing needs a few mods to Device, and one more parameter to persist. I'll add it in a new PR soon.

What I will add will work cross platform, except for the conversion of the encoded pairing URL into an QRcode image. I will add a function which will do that on macOS using a core image filter, but I don't know how you would do that on linux. If you were running from a terminal in macOS, you could write the image to a file and display it with 'open -a Preview.app image.png' or something like that. On linux I have no idea. I note that there are some javascript libraries on github which can create a QR code from block symbols, e.g https://github.com/gtanner/qrcode-terminal Perhaps you could translate that to swift if you needed that functionality.

PR #44, opened 2018-01-31
Auto Setup

gbrooker:
These changes add the capability to automatically pair from the iOS Home app or Camera app.
The setupCode and a new setupKey are moved to the persistent Configuration, for preservation over restarts. A setup hash is computed and added to the configuration record broadcast over MDNS. 

HomeKit matches the setup hash when performing an auto setup. A QR code should be displayed to the user representing a X-HM URI, which is an encoding of the setupCode, the primary accessory type and the setupKey. A new class is added which will create an NSImage from a string 

    QRcode(device.setupURI).asNSImage

The NSImage can then be added to an NSImageView. The QR Image generation is only available on macOS. Functions providing alternative representations such as an ASCII QR code could be added for other platforms at a later date by other authors.

When automatically pairing, the setupCode is not entered manually, so it does not need to be easily memorable, so can be auto generated (and preserved in the configuration).  I have therefore removed it as a required parameter to Device.init(). 

I have added an optional defaultSetupCode parameter to allow a caller to provide a preferred default setupCode if they really wanted to, instead of relying on an auto generated one. It will be ignored if a setupCode is found in the configuration. I think this is a kludge, and so I thought I'd discuss it with you before firming it up. I've not modified the test cases yet for that reason. I will update the test cases  once we've agreed on the best method of handling a user specified setupCode.

I just noticed that I left a few print() statements in Device.setupURI, I'll remove them in my next commit, once you've had a chance to review and try it out.


gbrooker:
One other note, the pairing identifier must be uppercase for this to work, this is fixed in the commit.

gbrooker:
For instance, a QR code as a String implementation could be inspired by :  
https://github.com/gtanner/qrcode-terminal

gbrooker:
or http://www.linux-magazine.com/Online/Features/Generating-QR-Codes-in-Linux

Sources/HAP/Server/Configuration.swift
```diff
@@ -7,25 +7,48 @@ fileprivate let logger = getLogger("hap.device")
...
+    let setupCode = String(format: "%03ld-%02ld-%03ld", n1, n2, n3)
+    
+    if !Device.isValid(setupCode: setupCode) {
+        return generateSetupCode()
+    }
+    return setupCode
+}
+
+// Generate a random four character setup key, used in setupURI and setupHash
+func generateSetupKey() -> String {
```
Bouke: Maybe we should change all these functions into static functions on the Device?

The first version of Swift used top-level functions quite a lot. Later versions reversed that decision. I still mix&match both styles, especially when those functions are somewhat specific to that file, but don't feel quite right being a class member.

Sources/HAP/Server/Device.swift
```diff
@@ -77,14 +80,14 @@ public class Device {
     ///   - accessories: accessories to be bridged
     convenience public init(
         bridgeInfo: Service.Info,
-        setupCode: String,
         storage: Storage,
-        accessories: [Accessory]) {
+        accessories: [Accessory],
+        defaultSetupCode: String = "automatic") {
```
Bouke: I don't like having magic strings; we could have an enum with an associated value.

```
enum SetupCode {
    case automatic
    case predefined(String)
}

init(setupCode: SetupCode = .automatic)
```

Sources/HAP/Server/Configuration.swift
```diff
@@ -7,25 +7,48 @@ fileprivate let logger = getLogger("hap.device")
...
+    let setupKey: String
     let publicKey: PublicKey
     let privateKey: PrivateKey
 
     /// Initializes a new configuration
-    init() {
+    init(defaultSetupCode: String = "automatic") {
         identifier = generateIdentifier()
+        setupCode = defaultSetupCode != "automatic" ? defaultSetupCode : generateSetupCode()
+        setupKey = generateSetupKey()
```
Bouke: Why do we need this on the configuration? Do we need to restore both the code and key when restarting the device? Or can we regenerate them every time?

Sources/HAP/Utils/QRCode.swift
```diff
@@ -0,0 +1,51 @@
+//
+//  QRCode.swift
+//  HAP
+//
+//  Created by Guy Brooker on 31/01/2018.
+//
```
Bouke: Please remove this file header comment

Bouke:
Just tested this myself, I must admit that this is REALLY COOL. I generated the QR code using some online tool. It would be very nice to generate the QR code in the terminal, but we'd need some ascii QR generated then.

Sources/HAP/Server/Configuration.swift
gbrooker: I agree, I am not a fan of putting anything in global scope, and had originally put these as static in Device, but for this PR I followed your lead with the generateIdentifier() function. I suggest putting this whole class as an extension of Device. The class will never be used outside of Device.

Sources/HAP/Server/Configuration.swift
gbrooker: I don't have access to the full spec, only what is made public, but I see in implementations such as Apples HomeKit Accessory Simulator and homebridge that these are preserved.

Sources/HAP/Utils/QRCode.swift
gbrooker: Sorry, that is the default header from XCode. I'll remove. Would be nice to get a credit in the README for my contributions for Dynamic accessory management and auto-setup.

gbrooker:
You could do, or just print the X-HM: URI to the terminal, and paste that text into any QR code generator. I use HAP from an app, so display a QR code in an NSImageView.

Alternatively you could port the javascript terminal QR code generator on github into Swift and add it to the QRCode class in a function 'asString"

gbrooker:
I've added some commits to address your comments. If the Device.init() api is OK like this, I'll update the test cases

Sources/HAP/Utils/QRCode.swift
Bouke: Sure! Feel free to add your name to the README!

Sources/HAP/Server/Device.swift
```diff
@@ -77,14 +81,14 @@ public class Device {
     ///   - accessories: accessories to be bridged
     convenience public init(
         bridgeInfo: Service.Info,
-        setupCode: String,
         storage: Storage,
-        accessories: [Accessory]) {
+        accessories: [Accessory],
+        setupCode: SetupCode = .automatic) {
```
Bouke: Can we move the argument to where it was? That means the API doesn't break as much.

Sources/HAP/Server/Device.SetupCode.swift
```diff
@@ -0,0 +1,67 @@
...
+
+extension Device {
+    public enum SetupCode:Equatable {
+        case automatic
+        case `default`(String)
+        
+        // HAP Specification lists certain setup codes as invalid
+        public var isValid: Bool {
+            switch (self) {
+            case .default(let setupCode):
```
Bouke: Maybe we could change `automatic` into `random` and `default` into `predefined`? I think that would make the names more straight-forward.

Sources/HAP/Server/Configuration.swift
```diff
@@ -4,55 +4,68 @@ import func Evergreen.getLogger
...
+    }
+
+    // Generate a random four character setup key, used in setupURI and setupHash
+    static internal func generateSetupKey() -> String {
+        return String(arc4random_uniform(1679616), radix:36, uppercase: true)
     }
-}
 
-typealias PrivateKey = Data
+    typealias PrivateKey = Data
```
Bouke: I'm not a fan of scoping this inside Device. The PrivateKey might be useful elsewhere, so the definition should probably move outside of this file. I'm fine with leaving it here for now.

Sources/HAP/Server/Device.swift
```diff
@@ -124,7 +129,7 @@ public class Device {
             configuration = try decoder.decode(Configuration.self, from: configData)
```
Bouke: So if the setupCode changed from `.automatic` to `.default`, we would effectively be ignoring that here. That might be confusing for users. What would be the best solution here to handle situations like this, also the other way around? Does the pairing / auto-setup break if we do change the `setupCode`?

Sources/HAP/Server/Device.swift
```diff
@@ -30,11 +30,15 @@ struct Box<T: Any>: Hashable, Equatable {
     }
 }
 
+
 public class Device {
     public let name: String
-    public let setupCode: String
     public let isBridge: Bool
-
+
```
Bouke: You can configure Xcode to automatically trim trailing whitespace, that would result in less whitespace changes.

Sources/HAP/Server/Device.SetupCode.swift
gbrooker: sure random is good, I don't like pre-defined because to me it implies that it would override what is in the configuration, which is not the case. Any other suggestion ?

Sources/HAP/Server/Configuration.swift
gbrooker: Ok, I'll revert Configuration back to a standalone class

Sources/HAP/Server/Device.swift
gbrooker: That's why I felt my solution was a bit of a kludge. Pairing won't break, if the code is changed. The hash code and X-HM will modify accordingly. What I could do is make the configuration store "lastSetupCode", and override that with one supplied by a user if specified. Your thoughts ?

Sources/HAP/Server/Device.swift
gbrooker: OK, I'll look for that

Sources/HAP/Server/Device.swift
gbrooker: OK

Sources/HAP/Server/Configuration.swift
gbrooker: Re-reading your comment, I guess you just mean moving PrivateKey out of Device, will do that

Sources/HAP/Server/Device.swift
gbrooker: I have reorganised things a little. If a `.predefined(setupCode)` is given as an argument, it will override whatever is in the configuration. I like this better. I still don't like the SetupCode enum names. I think I would prefer `.automatic` and `.override(setupCode)`, but I've kept your suggestion.

Bouke: I’m fine with override as well!

gbrooker:
I've addressed the comments above, fixed swiftlint pedantries and verified HAP suite tests. The travis tests failed due to arc4random_uniform not being available on that platform. I have provided an implementation which passes the travis tests, however I have no way to test it. Appreciate if you could verify that it functions correctly.

Sources/HAP/Server/Device.swift
gbrooker: OK, just refactored

Bouke:
> The travis tests failed due to arc4random_uniform not being available on that platform. I have provided an implementation which passes the travis tests, however I have no way to test it. Appreciate if you could verify that it functions correctly.

There's `Cryptor.Random.generateBytes` which we probably should use instead.

Bouke:
See my two commits to the branch and let me know what you think.

Bouke:
Furthermore I think it'll be good to merge!

gbrooker:
I really love the ExpressibleByStringLiteral, that is so elegant and impressive. I never knew that protocol existed. I've learnt something today, thanks.

However, I really don't like the changes to random number. Why replace arc4random_uniform on macOS ? Posts I have read say that function will be implemented in Foundation on linux eventually, so surely it is the right method to use, replaced by a temporary workaround on that platform for now. Why replace the recursion with a loop ? Why loop for 100 times ? The number is quite arbitrary, and the function would return an invalid code if it ever got past 100 tries. I much prefer my version. How about a compromise using Random.generate(byteCount) on linux only ?

Bouke:
> I really love the ExpressibleByStringLiteral, that is so elegant and impressive. I never knew that protocol existed. I've learnt something today, thanks.

Yeah it's really nice. Don't overuse it, however, as it might result in the compiler having to resolve multiple branches.

> However, I really don't like the changes to random number. Why replace arc4random_uniform on macOS ? Posts I have read say that function will be implemented in Foundation on linux eventually, so surely it is the right method to use, replaced by a temporary workaround on that platform for now. 
> How about a compromise using Random.generate(byteCount) on linux only ?

Fair enough. I'll write up a different implementation.

> Why replace the recursion with a loop ? Why loop for 100 times ? The number is quite arbitrary, and the function would return an invalid code if it ever got past 100 tries. I much prefer my version. 

The function would not return an invalid code if it doesn't find a valid setup code, but traps instead. I replaced the recursion to put an upper limit on the amount of tries that we do. I haven't calculated the chance, but I think we can safely assume that we should find a setup code in 100 tries. If we can't find a setup code in 100 tries, it's more probable that we have an issue with the random number generator, which would otherwise result in indefinite recursion.

Bouke:
I've written a new implementation of `arc4random_uniform` for Linux and guarded against modulo bias and also replaced the loop for an indefinite loop. What do you think of it?

Bouke:
I've added a QRCode generator library and a transformation to ASCII. It displays nicely within the Xcode debug console, however the boxes show incorrectly in iTerm however.

gbrooker:
Great additions, you've been busy yesterday ! I struggled at first incorporating the CQRCode framework into my app, Xcode kept giving a build error about a missing module, even though the framework was in the LinkedFrameworks list. I finally managed to fix that by adding a line to Header Search Path in build setting with `$(SRCROOT)/../HAP/HAP.xcodeproj/GeneratedModuleMap/CQRCode` but I wonder if there is a better way. Unless I'm doing something wrong, perhaps it would be useful for people to include a mention in the documentation.

I made one small change. Thanks for spotting the need to pad the setupKey, but I moved that to the `generateSetupKey()` function, so it is corrected at source.

I added a couple of other QR representations, which may work better in some terminals. `asBigText` will use two blocks per pixel, and `asAscii` is similar but uses two `#` characters if only a very basic character set is available. These both work for me, though the ascii representation needed me to step back quite a bit from the screen.

gbrooker:
Thanks for the changes to arc4random, I think the result is much better.

Bouke:
> Great additions, you've been busy yesterday !

Thanks!

> I struggled at first incorporating the CQRCode framework into my app, Xcode kept giving a build error about a missing module, even though the framework was in the LinkedFrameworks list. I finally managed to fix that by adding a line to Header Search Path in build setting with `$(SRCROOT)/../HAP/HAP.xcodeproj/GeneratedModuleMap/CQRCode` but I wonder if there is a better way. Unless I'm doing something wrong, perhaps it would be useful for people to include a mention in the documentation.

Using SwiftPM you can do `swift package generate-xcodeproj` and that'll setup the header search path correctly. In your case you're doing a Cocoa-app, I'm not quite sure how that would nicely fit in here -- are you using a workspace?

> I tried both using only one ascii character per pixel, but as characters as not square, it resulted in a tall thin QR code that I could only get recognised when holding the phone at acute angles from the screen.

Yeah I ran into the same problem. That's why I combined two lines into a single character in the first place.

By the way; your git committer name/email is off due to the curly quotes. You might want to verify that you've set them correctly: `git config --global user.name` and `git config --global user.email`.

I've merged the PR, great work on this nice addition!

PR #48, opened 2018-02-04
custom service type

tgu:
By duplicating the code we can get custom service types, a little bit ugly, especially if we also want custom accessory types. Should find a way to share code and still by type-safe.


Bouke:
👍

PR #50, opened 2018-02-05
Pairing observer

gbrooker:
The last commit to HAP/master removed the `onConfigurationChange` callback in `Device`. This was useful for monitoring pairing status by the application calling HAP. 

This pull request proposes to add back a callback mechanism for specifically monitoring pairing status. An enum communicating the pairing stage `PairingEvent` is passed to the callback.

### DIFF
--- a/Sources/HAP/Controllers/PairSetupController.swift
+++ b/Sources/HAP/Controllers/PairSetupController.swift
@@ -55,7 +55,8 @@ class PairSetupController {
         }
 
         // Notify listeners of the pairing event and record the paring state
-        device.changePairingState(.pairing)
+        // swiftlint:disable:next force_try
+        try! device.changePairingState(.pairing)
 
         let (salt, serverPublicKey) = session.server.getChallenge()
 
@@ -74,7 +75,6 @@ class PairSetupController {
     func verifyRequest(_ data: PairTagTLV8, _ session: Session) throws -> PairTagTLV8? {
         guard let clientPublicKey = data[.publicKey], let clientKeyProof = data[.proof] else {
             logger.warning("Invalid parameters")
-            device.changePairingState(.notPaired)
             throw Error.invalidSetupState
         }
 
@@ -85,7 +85,6 @@ class PairSetupController {
                                                                      keyProof: clientKeyProof)
             else {
                 logger.warning("Invalid PIN")
-                device.changePairingState(.notPaired)
                 throw Error.authenticationFailed
         }
 
@@ -112,12 +111,10 @@ class PairSetupController {
         guard let plaintext = try? ChaCha20Poly1305.decrypt(cipher: encryptedData,
                                                             nonce: "PS-Msg05".data(using: .utf8)!,
                                                             key: encryptionKey) else {
-            device.changePairingState(.notPaired)
             throw Error.couldNotDecryptMessage
         }
 
         guard let data: PairTagTLV8 = try? decode(plaintext) else {
-            device.changePairingState(.notPaired)
             throw Error.couldNotDecodeMessage
         }
 
@@ -125,7 +122,6 @@ class PairSetupController {
             let username = data[.identifier],
             let signatureIn = data[.signature]
             else {
-                device.changePairingState(.notPaired)
                 throw Error.invalidParameters
         }
 
@@ -141,12 +137,7 @@ class PairSetupController {
                      username +
                      publicKey
 
-        do {
-            try Ed25519.verify(publicKey: publicKey, message: hashIn, signature: signatureIn)
-        } catch {
-            device.changePairingState(.notPaired)
-            throw error
-        }
+        try Ed25519.verify(publicKey: publicKey, message: hashIn, signature: signatureIn)
 
         let hashOut = deriveKey(algorithm: .sha512,
                                 seed: session.server.sessionKey!,
@@ -157,7 +148,6 @@ class PairSetupController {
             device.publicKey
 
         guard let signatureOut = try? Ed25519.sign(privateKey: device.privateKey, message: hashOut) else {
-            device.changePairingState(.notPaired)
             throw Error.couldNotSign
         }
 
@@ -176,7 +166,6 @@ class PairSetupController {
                                                                        nonce: "PS-Msg06".data(using: .utf8)!,
                                                                        key: encryptionKey)
             else {
-                device.changePairingState(.notPaired)
                 throw Error.couldNotEncrypt
         }
 
@@ -188,10 +177,5 @@ class PairSetupController {
             (.encryptedData, encryptedResultInner)
         ]
         return resultOuter
-    }
-
-    func unknownRequest(_ data: PairTagTLV8) throws -> PairTagTLV8 {
-        device.changePairingState(.notPaired)
-        throw Error.invalidParameters
     }
 }
--- a/Sources/HAP/Controllers/PairSetupController.swift
+++ b/Sources/HAP/Controllers/PairSetupController.swift
@@ -16,6 +16,7 @@ class PairSetupController {
         case couldNotDecodeMessage
         case couldNotSign
         case couldNotEncrypt
+        case errorWithResponse(PairTagTLV8)
     }
 
     let device: Device
@@ -32,6 +33,33 @@ class PairSetupController {
             throw Error.invalidPairingMethod
         }
 
+        // If the accessory is already paired it must respond with
+        // Error_Unavailable
+        if device.state == .paired {
+            throw Error.errorWithResponse([
+                (.state, Data(bytes: [PairSetupStep.startResponse.rawValue])),
+                (.error, Data(bytes: [PairError.unavailable.rawValue]))
+            ])
+        }
+
+        // If the accessory has received more than 100 unsuccessful
+        // authentication attempts it must respond with
+        // Error_MaxTries
+        // TODO
+
+        // If the accessory is currently performing a Pair Setup operation with
+        // a different controller it must respond with
+        // Error_Busy
+        if device.state == .pairing {
+            throw Error.errorWithResponse([
+                (.state, Data(bytes: [PairSetupStep.startResponse.rawValue])),
+                (.error, Data(bytes: [PairError.busy.rawValue]))
+            ])
+        }
+
+        // Notify listeners of the pairing event and record the paring state
+        device.changePairingState(.pairing)
+
         let (salt, serverPublicKey) = session.server.getChallenge()
 
         logger.info("Pair setup started")
@@ -39,17 +67,21 @@ class PairSetupController {
         logger.debug("<-- B \(serverPublicKey.hex)")
 
         let result: PairTagTLV8 = [
-            .state: Data(bytes: [PairSetupStep.startResponse.rawValue]),
-            .publicKey: serverPublicKey,
-            .salt: salt
+            (.state, Data(bytes: [PairSetupStep.startResponse.rawValue])),
+            (.publicKey, serverPublicKey),
+            (.salt, salt)
         ]
         return result
     }
 
-    func verifyRequest(_ data: PairTagTLV8, _ session: Session) -> PairTagTLV8? {
+    func verifyRequest(_ data: PairTagTLV8, _ session: Session) throws -> PairTagTLV8? {
         guard let clientPublicKey = data[.publicKey], let clientKeyProof = data[.proof] else {
             logger.warning("Invalid parameters")
-            return nil
+            device.changePairingState(.notPaired)
+            throw Error.errorWithResponse([
+                (.state, Data(bytes: [PairSetupStep.verifyResponse.rawValue])),
+                (.error, Data(bytes: [PairError.unknown.rawValue]))
+            ])
         }
 
         logger.debug("--> A \(clientPublicKey.hex)")
@@ -59,18 +91,18 @@ class PairSetupController {
                                                                      keyProof: clientKeyProof)
             else {
                 logger.warning("Invalid PIN")
-                let result: PairTagTLV8 = [
-                    .state: Data(bytes: [PairSetupStep.verifyResponse.rawValue]),
-                    .error: Data(bytes: [PairError.authenticationFailed.rawValue])
-                ]
-                return result
+                device.changePairingState(.notPaired)
+                throw Error.errorWithResponse([
+                    (.state, Data(bytes: [PairSetupStep.verifyResponse.rawValue])),
+                    (.error, Data(bytes: [PairError.authenticationFailed.rawValue]))
+                ])
         }
 
         logger.debug("<-- HAMK \(serverKeyProof.hex)")
 
         let result: PairTagTLV8 = [
-            .state: Data(bytes: [PairSetupStep.verifyResponse.rawValue]),
-            .proof: serverKeyProof
+            (.state, Data(bytes: [PairSetupStep.verifyResponse.rawValue])),
+            (.proof, serverKeyProof)
         ]
         return result
     }
@@ -89,10 +121,12 @@ class PairSetupController {
         guard let plaintext = try? ChaCha20Poly1305.decrypt(cipher: encryptedData,
                                                             nonce: "PS-Msg05".data(using: .utf8)!,
                                                             key: encryptionKey) else {
-                                                                throw Error.couldNotDecryptMessage
+            device.changePairingState(.notPaired)
+            throw Error.couldNotDecryptMessage
         }
 
         guard let data: PairTagTLV8 = try? decode(plaintext) else {
+            device.changePairingState(.notPaired)
             throw Error.couldNotDecodeMessage
         }
 
@@ -100,6 +134,7 @@ class PairSetupController {
             let username = data[.identifier],
             let signatureIn = data[.signature]
             else {
+                device.changePairingState(.notPaired)
                 throw Error.invalidParameters
         }
 
@@ -115,10 +150,12 @@ class PairSetupController {
                      username +
                      publicKey
 
-        try Ed25519.verify(publicKey: publicKey, message: hashIn, signature: signatureIn)
-
-        // At this point, the pairing has completed. The first controller is granted admin role.
-        device.add(pairing: Pairing(identifier: username, publicKey: publicKey, role: .admin))
+        do {
+            try Ed25519.verify(publicKey: publicKey, message: hashIn, signature: signatureIn)
+        } catch {
+            device.changePairingState(.notPaired)
+            throw error
+        }
 
         let hashOut = deriveKey(algorithm: .sha512,
                                 seed: session.server.sessionKey!,
@@ -129,13 +166,14 @@ class PairSetupController {
             device.publicKey
 
         guard let signatureOut = try? Ed25519.sign(privateKey: device.privateKey, message: hashOut) else {
+            device.changePairingState(.notPaired)
             throw Error.couldNotSign
         }
 
         let resultInner: PairTagTLV8 = [
-            .identifier: device.identifier.data(using: .utf8)!,
-            .publicKey: device.publicKey,
-            .signature: signatureOut
+            (.identifier, device.identifier.data(using: .utf8)!),
+            (.publicKey, device.publicKey),
+            (.signature, signatureOut)
         ]
 
         logger.debug("<-- identifier \(self.device.identifier)")
@@ -147,13 +185,25 @@ class PairSetupController {
                                                                        nonce: "PS-Msg06".data(using: .utf8)!,
                                                                        key: encryptionKey)
             else {
+                device.changePairingState(.notPaired)
                 throw Error.couldNotEncrypt
         }
 
+        // At this point, the pairing has completed. The first controller is granted admin role.
+        device.add(pairing: Pairing(identifier: username, publicKey: publicKey, role: .admin))
+
         let resultOuter: PairTagTLV8 = [
-            .state: Data(bytes: [PairSetupStep.keyExchangeResponse.rawValue]),
-            .encryptedData: encryptedResultInner
+            (.state, Data(bytes: [PairSetupStep.keyExchangeResponse.rawValue])),
+            (.encryptedData, encryptedResultInner)
         ]
         return resultOuter
+    }
+
+    func unknownRequest(_ data: PairTagTLV8) throws -> PairTagTLV8 {
+        device.changePairingState(.notPaired)
+        throw Error.errorWithResponse([
+            (.state, Data(bytes: [PairSetupStep.waiting.rawValue])),
+            (.error, Data(bytes: [PairError.unknown.rawValue]))
+        ])
     }
 }
--- a/Sources/HAP/Controllers/PairSetupController.swift
+++ b/Sources/HAP/Controllers/PairSetupController.swift
@@ -16,7 +16,10 @@ class PairSetupController {
         case couldNotDecodeMessage
         case couldNotSign
         case couldNotEncrypt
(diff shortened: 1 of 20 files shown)


Sources/HAP/Endpoints/pairSetup().swift
```diff
@@ -56,6 +57,7 @@ func pairSetup(device: Device) -> Application {
             case .keyExchangeRequest:
                 let session = try getSession(connection)
                 response = try controller.keyExchangeRequest(data, session)
+                _ = device.onPairingEvent.map { $0(device, .pairingCompleted) }
```
Bouke: The response could also be an error at this point. We shouldn't hook into these functions, but do that on the actual `Device` instance. These callbacks should be fired from `Device.add(pairing:)` and `Device.remove(pairingWithIdentifier:)` instead.

Bouke:
> The last commit to HAP/master removed the onConfigurationChange callback in Device. This was useful for monitoring pairing status by the application calling HAP.

Yes I removed it because the package internally relied upon the callback being used for updating the bonjour records. So if the callbacks were emptied (public setter), the bonjour records wouldn't be updated anymore.

gbrooker:
I was going to propose to add back a simple call in `notifyConfigurationChange` in `Device`, but chose to go deeper into the pairing process, because I wanted a notification when the pairing actually started, well before `add(pairing:)` is called. I don't see a way to achieve that without hooking into the endpoint functions. Perhaps it would be safer if I created a `notifyPairingState` function, which posted the callbacks on the main queue, which would protect the endpoints from a listener delaying or upsetting the pairing process.

Bouke:
> I was going to propose to add back a simple call in notifyConfigurationChange in Device, but chose to go deeper into the pairing process, because I wanted a notification when the pairing actually started, well before add(pairing:) is called. I don't see a way to achieve that without hooking into the endpoint functions. 

I don't like having those callbacks inside the endpoints, but there currently also isn't another way of hooking those events. The problem is mostly around the "pairing failed" event, which might also need to trigger in these situations:

* failed to complete the pairing process (timeout) -- currently not detected
* wrong pairing code provided -- handled elsewhere in the controller
* incorrect parameters provided -- the controller doesn't always throw but returns a `.badRequest`

How about we only implement the following pairing events for now:

* new pairing
* removed pairing

We can revisit the design of the end points and controllers separately.

> Perhaps it would be safer if I created a notifyPairingState function, which posted the callbacks on the main queue, which would protect the endpoints from a listener delaying or upsetting the pairing process.

The endpoints are already executed on the main thread, I think. However it would be better to decouple, so the callbacks don't delay the pairing process.

gbrooker:
How about this. I've added a timeout to the pairing process, and protected the `pairSetup()` endpoint from possible delays in the listener callbacks by calling them asynchronously on the main queue. I have removed the notification from the `pairings()` callback, so notifications are only made from `pairSetup()` and `Device`. I think this captures all the interesting states and failure modes of the process. What do you think ?

gbrooker:
I think I have managed to sort out the mess with my fork synchronisation. The Fork.app was extremely useful to understand the mess I had got in (thanks a lot for that recomendation). I had four separate branches cross merging in different directions somehow creating a GitHub spaghetti bolognese. I believe that my Github master is now properly synchronised with yours, and this pull request branch has been rebased on the HEAD and squashed to one commit, so hopefully easier to review.

Bouke:
> I think I have managed to sort out the mess with my fork synchronisation. The Fork.app was extremely useful to understand the mess I had got in (thanks a lot for that recomendation). I had four separate branches cross merging in different directions somehow creating a GitHub spaghetti bolognese. I believe that my Github master is now properly synchronised with yours, and this pull request branch has been rebased on the HEAD and squashed to one commit, so hopefully easier to review.

Yeah the history of this commit is looking much better now. The git author name of the latest commit is still wrong, but could be due to you rebasing the commits.

Bouke:
> How about this. I've added a timeout to the pairing process, and protected the pairSetup() endpoint from possible delays in the listener callbacks by calling them asynchronously on the main queue. I have removed the notification from the pairings() callback, so pairing notifications are only made from pairSetup() and from Device.

The code is assuming a single pairing controller at a time, which introduces a race condition.
So when there are two simultaneous pairing controllers, unexpected pairing events might be emitted. I'm still not convinced of having these pairing events in the first place, unless there's an elegant implementation that covers all situations.

gbrooker:
Interesting thought, but the HAP spec  (4.7.2-3) says if the accessory is currently in a pairing process it should respond with `PairError.busy` and if it is already paired it should respond with `PairError.unavailable`. Unless I missed it, these conditions don't seem to be tested and rejected accordingly. If they were, surely there wouldn't be a possibility for a race condition and unexpected pairing events. As far as I can see, the unlikely scenario of multiple Bridges/Devices pairing simultaneously on the same server would not interfere with each others pairing process.

I'll take a look at incorporating those conditions from HAP spec 4.7.2.

Bouke:
> Interesting thought, but the HAP spec (4.7.2-3) says if the accessory is currently in a pairing process it should respond with PairError.busy and if it is already paired it should respond with PairError.unavailable. Unless I missed it, these conditions don't seem to be tested and rejected accordingly. 

TIL. The code for setup pairing could be somewhat simplified knowing this. I still would want to isolate the pairing within a controller for testability. However some assumptions could be made with regards to pair events as well.

> If they were, surely there wouldn't be a possibility for a race condition and unexpected pairing events. As far as I can see, the unlikely scenario of multiple Bridges/Devices pairing simultaneously on the same server would not interfere with each others pairing process.

Multiple bridges would mean multiple servers, which would also mean additional instances of those pair setup endpoints. So I think we should be fine, as long as we don't introduce library global state.

gbrooker:
I have migrated the pairing notification code to the PairingController, and have added tests for various requirements from the spec, and returned appropriate error codes (e.g `busy` or `unavailable`). If an error is returned in the TLV8, then the web response will also report an error as specified in the spec. I note that there are a number of errors thrown by the PairController code which result in an empty TLV8 response, but I would suggest these should be returned as proper TLV8 errors. What are your thoughts ?

As required by the spec, I have also added support for auto renaming the device (e.g `Bridge (2)`) if its requested name collides with an existing server on the subnet. The name field in the TXT record is updated to reflect the rename. Renaming is now the default behaviour, but In some cases an app may wish to handle this directly, so I have added a mechanism for the app to be notified of the name collision.

gbrooker:
So how about this last update ? Can we merge ?

Bouke:
I've been busy the past few days, I didn't get around to look into this, sorry. While I haven't combed through the code in detail, below two major things I think we need to resolve.

I think it would be better to formalise the states that the device can be in using an enum. The various states that I imagine we would have are these; along with the possible transitions between them:


[edit diagram](https://www.planttext.com/?text=SoWkIImgAStDuIh9BCb9LV1m351mIItBJ4vL24ZCBChCIrS8BaaiIUMArefLqDMrK_3BBm59fAPmmLbWSQXgBfWk936tnCmSe7eK7ffPI0QX2Y8R1HHoICrB0JeM0000)

Regarding the auto-renaming, I was under the impression that `NSNetService` provides this functionality by default already. Why would we want to opt-out of that functionality? And if `NSNetService` did rename our service, do we want to change our advertisement or keep the user-provided service name?

gbrooker:
Hi Bouke, no worries. 

Regarding the pairing states, yes I can change `PairingEvent` into a `PairingState`, but there would only be three states: `Paired`, `Unpaired` and `Pairing`. PairingError is just an event transitioning from `Pairing` to `Unpaired` states. Let me know if you want me to proceed in that direction.

Regarding auto-renaming, the current master branch of HAP fails silently when creating a server with the same name as another HAP server on the same subnet. i.e Server thinks it is advertising itself, but no mDNS record is published for the second instance. In order to get `NSNetService` to auto rename as required by the spec, we need to call `service.publish()` with the `options:` parameter explicitly set to no flags (fixed in this PR). It appears that on macOS the default behaviour of `publish()` is not to auto-rename. If `NSNetService` renames the mDNS entry, HAP needs to modify the device name in the Info record (and hence the TXT record) accordingly (spec 5.4). If the Info record was not changed,  it would not be possible for a user to distinguish between the devices in the iOS Home app (this is also taken care of in this PR).

In some cases (such as the apps I am working on) it only makes sense to have one Bridge of its type on the sub-net, so the existence of another Bridge with the same name should be fed back to the caller to handle accordingly. I have added a parameter to `Device.init()` called `onNameCollision:` which passes a function to be called if mDNS detects a name collision when publishing. If this parameter is supplied, mDNS auto-renaming is disabled. I note that it is not possible to make `Device.init()` fail or throw an exception, because a collision will only be notified asynchronously to Server.

gbrooker:
Just added an implementation of listPairings. It required an alternative Array implementation of PairTagTLV8, to preserve the exact order of the TLV8 elements. In fact I'm not sure how the existing implementation can guarantee the right order of elements if there is more than one pair tag in the response...

I'm not sure how to test this, when does a HAP controller actually ask for the list of pairings ?

It also includes the fix I mentioned above for the Info record when auto-renamed (I thought it was in the previous commit).

Bouke:
Thanks for your comments and changes. It's taken me a little while to get back to you, but hopefully we can get your PR merged soon.

> Regarding the pairing states, yes I can change PairingEvent into a PairingState, but there would only be three states: Paired, Unpaired and Pairing. PairingError is just an event transitioning from Pairing to Unpaired states. Let me know if you want me to proceed in that direction.

PairingError would be a transient state; if the Device enters that state, it would immediately transition into the Unpaired state. The reason for the existence of this state is that the controller of the Device would receive a notification that the Device entered that state and can handle that accordingly. Similar to what you proposed with the event nofications, such that you can provide feedback through some sort of UI.

> In order to get NSNetService to auto rename as required by the HAP spec, we need to call service.publish() with the options: parameter explicitly set to no flags (fixed in this PR).

That is very strange, from the [docs](https://developer.apple.com/documentation/foundation/netservice/1416480-publish):

> This method returns immediately, with success or failure indicated by the callbacks to the delegate. This is equivalent to calling `publish(options:)` with the default options (0).

> If NSNetService renames the mDNS entry, HAP needs to modify the device name in the Info record (and hence the TXT record) accordingly (spec 5.4). If the Info record was not changed, it would not be possible for a user to distinguish between the devices in the iOS Home app (this is also taken care of in this PR).

I'd like for this package to support auto renaming, but I fail to see the need for disabling auto renaming.

> Just added an implementation of listPairings. It required an alternative Array implementation of PairTagTLV8, to preserve the exact order of the TLV8 elements. In fact I'm not sure how the existing implementation can guarantee the right order of elements if there is more than one pair tag in the response...

The current implementation is modeled like a dictionary. There can only be a single instance of a certain key. It seems that this is an oversimplification and should've been an array instead, like what you've created.

I think that the modeling of the dictionary should be removed and replaced with your array implementation throughout. But I'm fine with leaving things as-is for now.

> I'm not sure how to test this, when does a HAP controller actually ask for the list of pairings ?

I ran some tests a while back and I remember that I couldn't get it to ever ask for the list of pairings. So I'm not sure if it is in use even.

gbrooker:
Hi Bouke, nice to hear from you. 

I'll have a think about the `PairingStates`, and upload some changes in the coming days. 

NSNetService: it seems that Apple's documentation for `NetService.publish()` is either wrong or misleading. Under macOS when called with no parameters it is apparently equivalent to `NetService.publish(options: .noAutoRename)`. When passed the options parameter `NetService.Options(rawValue: 0)` auto renaming functions as expected on macOS. 

In my own use case I am implementing a HAP server which bridges to all discoverable devices on the subnet which use a certain protocol. There should only ever be one instance of such a bridge for any subnet, otherwise multiple bridges would be competing to control the same devices. Therefore if the bridge name collides with another mDNS record with the same name, it needs to inform the user that a duplicate exists and gracefully exit, rather than auto rename. 

As you suggest, I'll completely replace the dictionary implementation of `PairTagTVL8` with my array implementation.

I have a couple of other branches I've ben working on which I'll put into PR's soon, one to fix a race condition in Server.Connection and another to allow an app to monitor HAP get requests (enabling lazy updates)

Bouke:
> In my own use case I am implementing a HAP server which bridges to all discoverable devices on the subnet which use a certain protocol. There should only ever be one instance of such a bridge for any subnet, otherwise multiple bridges would be competing to control the same devices. Therefore if the bridge name collides with another mDNS record with the same name, it needs to inform the user that a duplicate exists and gracefully exit, rather than auto rename.

I feel that this doesn't fit with the scope of this library. For example, does the inability of publishing the service mean that there's already an instance running of your app? I don't think that we should be adding such complexity to this library.

> I have a couple of other branches I've ben working on which I'll put into PR's soon, one to fix a race condition in Server.Connection and another to allow an app to monitor HAP get requests (enabling lazy updates)

Nice! The current connection handling is error prone and probably has additional threading (gcd) issues. It would be interesting to see what the recently released [swift-nio](https://github.com/apple/swift-nio) framework could bring to the table here.

gbrooker:
Thanks for pushing back on the mDNS name collision changes. You are right, it is untidy and better left out. Thankfully there is a simple workaround, as I added code to catch a net service rename and update the name of the Device accordingly (as required by the spec), a host app only needs to watch for changes to the `Service.Info.name` characteristic.

gbrooker:
There seem to be some issues with travis on GitHub today...

Sources/HAP/Controllers/PairSetupController.swift
```diff
@@ -32,6 +56,34 @@ class PairSetupController {
...
+        // If the accessory has received more than 100 unsuccessful
+        // authentication attempts it must respond with
+        // Error_MaxTries
+        // TODO
+
+        // If the accessory is currently performing a Pair Setup operation with
+        // a different controller it must respond with
+        // Error_Busy
+        if device.lastPairingEvent == .pairingStarted ||
+            device.lastPairingEvent == .pairingVerified {
```
Bouke: Having a state machine, we wouldn't have to have all such logic throughout various controllers. We would ask the state machine if we can change into "pairing started", which would fail if we're not in the correct state (see also state diagram).

Bouke:
> There seem to be some issues with travis on GitHub today...

Yeah they seem to be resolved now.

gbrooker:
I've replaced all uses of the `PairTagTLV8` dictionary implementation entirely with the array implementation.

I still need to see what changes I would have to make to implement the state machine you have in mind. I not 100% convinced it would bring much benefit, but I'll explore the idea and try it out.

Indeed swiftnio looks quite interesting. I'm very happy to see Promises embraced by Apple, though I would have liked it in Foundation. It looks like it could provide most of what HAP needs to communicate with the controllers, replacing Kitura and BlueSockets, but would mean rewriting the endpoints and controllers. I'm slightly off-put by the replacement of some core Foundation data types such as the nio EventLoops replacing GCD DispatchQueues and ByteBuffer replacing Data. Have you experimented with it ?

Sources/HAP/Endpoints/pairSetup().swift
```diff
@@ -38,36 +38,48 @@ func pairSetup(device: Device) -> Application {
...
         } catch {
             logger.warning(error)
             connection.context[SESSION_KEY] = nil
             response = nil
         }
         if let response = response {
-            return Response(status: .ok, data: encode(response), mimeType: "application/pairing+tlv8")
+            let status: Response.Status
+            if let errorData = response[.error] {
+                status = .badRequest
```
Bouke: When an invalid PIN is entered, iOS expects the HTTP status code to be 200 -- not 400 (bad request).

Bouke:
I've created PR 2 in your repo, showing a state machine based on your work: https://github.com/gbrooker/HAP/pull/2.

I don't quite like all the changes to the controllers. I've separated the end points from the controllers, such that the controllers don't need to aware of the network layer. The controllers are ideally separated from the device, so that the controllers can work independently and most importantly also be tested separately. There's still some coupling, but I'd rather not have even tighter coupling.

> Indeed swiftnio looks quite interesting. I'm very happy to see Promises embraced by Apple, though I would have liked it in Foundation. It looks like it could provide most of what HAP needs to communicate with the controllers, replacing Kitura and BlueSockets, but would mean rewriting the endpoints and controllers. I'm slightly off-put by the replacement of some core Foundation data types such as the nio EventLoops replacing GCD DispatchQueues and ByteBuffer replacing Data. Have you experimented with it ?

I agree; what they're basically saying is that Foundation can't be used for performance reasons. I would expect that higher-level frameworks would be built on top of swift-nio which might use Foundation instead. I haven't experienced with it yet. For HAP I would like to stick with high-level frameworks to keep the implementation tidy, not having to bother too much with low-level details.

Sources/HAP/Endpoints/pairSetup().swift
gbrooker: OK I've changed it to return 200 OK if there is a TVL8 response to send, as it was before

Sources/HAP/Controllers/PairSetupController.swift
```diff
@@ -23,7 +27,9 @@ class PairSetupController {
         self.device = device
     }
 
-    func startRequest(_ data: PairTagTLV8, _ session: Session) throws -> PairTagTLV8 {
+    var timeOut = true
```
Bouke: This variable is no longer in use.

Sources/HAP/Controllers/PairSetupController.swift
```diff
@@ -88,18 +127,22 @@ class PairSetupController {
 
         guard let plaintext = try? ChaCha20Poly1305.decrypt(cipher: encryptedData,
                                                             nonce: "PS-Msg05".data(using: .utf8)!,
-                                                            key: encryptionKey) else {
-                                                                throw Error.couldNotDecryptMessage
+                                                            key: encryptionKey)
+            else {
+            device.changePairingState(.notPaired)
+            throw Error.couldNotDecryptMessage
```
Bouke: These two lines could use an extra level of indentation for readability.

Sources/HAP/Controllers/PairSetupController.swift
```diff
@@ -32,24 +38,56 @@ class PairSetupController {
...
+        // Error_Busy
+        if device.state == .pairing {
+            return [
+                (.state, Data(bytes: [PairSetupStep.startResponse.rawValue])),
+                (.error, Data(bytes: [PairError.busy.rawValue]))
+            ]
+        }
+
+        // Notify listeners of the pairing event and record the paring state
+        device.changePairingState(.pairing)
```
Bouke: We don't reset the state after some timeout. This might leave the device in the pairing state, unable to be paired after some failure. I will have a look at how the controller can be decoupled from the device, and maybe rely on the session state that is being passed in. I will get back on this later.

Sources/HAP/Endpoints/pairSetup().swift
```diff
@@ -38,36 +38,44 @@ func pairSetup(device: Device) -> Application {
...
-                response = nil
+                response = try controller.unknownRequest(data)
             }
         } catch {
             logger.warning(error)
             connection.context[SESSION_KEY] = nil
             response = nil
         }
         if let response = response {
+            if let errorData = response[.error] {
```
Bouke: I rather stick with throwing errors from the controller, which can be caught in the endpoint and then converted to the various error responses as necessary.

Sources/HAP/Utils/TLV8.swift
```diff
@@ -1,15 +1,24 @@
 import Foundation
 
-typealias PairTagTLV8 = [PairTag: Data]
+typealias PairTagTLV8 = (PairTag, Data)
```
Bouke: This would keep it more in spirit of the original type and also results in less changes: `typealias PairTagTLV8 = [(PairTag, Data)]`

Tests/HAPTests/TLV8Tests.swift
```diff
@@ -14,12 +14,19 @@ class TLV8Tests: XCTestCase {
...
     func test() {
         let publicKey = Data(repeating: 0, count: 256) + Data(repeating: 1, count: 256) + Data(repeating: 2, count: 256)
-        let original: PairTagTLV8 = [
-            .publicKey: publicKey
+        let original: [PairTagTLV8] = [
+            (.publicKey, publicKey)
         ]
         let encoded = encode(original)
-        let decoded: PairTagTLV8 = try! decode(encoded)
-        XCTAssertEqual(original, decoded)
```
Bouke: Can we make the 'new' PairTagTLV8 equatable, what would be needed for that?

Sources/HAP/Controllers/PairSetupController.swift
gbrooker: Yes, thanks for picking that up

Sources/HAP/Controllers/PairSetupController.swift
gbrooker: Got it

Sources/HAP/Controllers/PairSetupController.swift
gbrooker: OK let me know if you need me to look at anything

Sources/HAP/Utils/TLV8.swift
gbrooker: I preferred to use the type PairTagTLV8 to refer to a single pair, and [PairTagTLV8] or PairTagTLV8Array to refer to a sequence of PairTag’s, however I will rename as you requested.

Tests/HAPTests/TLV8Tests.swift
gbrooker: Unfortunately Swift 4 does not permit tuples to implement `Equatable` (or any other Protocol), but it is possible to add an `==`  operator to the PairTagTVL8 array, which I’ve now implemented in TVL8.swift. Even though it has an equality operator, it still can’t conform to the `Equatable` protocol, but it does simplify the test case code to         
`XCTAssertTrue(original == decoded)`

Sources/HAP/Endpoints/pairSetup().swift
gbrooker: OK I've addressed that in the latest commit.

gbrooker:
I've moved things around a little in `Device.swift` to facilitate merging with my other PR which also introduces changes in that file.

gbrooker:
The linux travis builds were failing, as version 1.0.12 of libsodium is hardcoded in the `travis.yml` file, but is no longer available for download. I have updated `travis.yml` to use version 1.0.16, which is what is installed by brew on the macOS build.

gbrooker:
I have squashed the commits in this branch, and re-based to the latest master. I've then changed onPairingStateChange to use a didChangePairingState function in the new DeviceDelegate.

gbrooker:
Interesting, I initially implemented the error types in PairSetupController as you have done in 55ef409 but thought it was neater that all the TLV8 responses were constructed in the controller class, rather than have the errors constructed in the `pairSetup()` endpoint, and the others constructed in the controller. Both methods work fine.

Bouke:
> Interesting, I initially implemented the error types in PairSetupController as you have done in 55ef409 but thought it was neater that all the TLV8 responses were constructed in the controller class, rather than have the errors constructed in the pairSetup() endpoint, and the others constructed in the controller. Both methods work fine.

Yeah I'm not completely happy with the split between the controller and the endpoint. I'd like to have a clear separation to aid unit testing.

Bouke:
Note that your git committer name is corrupted again. You're committing with: (note the curly quotes)

    “Guy <“guy.brooker@me.com”>

Bouke:
I've merged in this branch with some adjustments in 2f048a3c32b8705491a558df5565f022bfd1a695 and 517ab9670929a685c1f0db6d8244c1430b77849a. Thanks!

PR #51, opened 2018-02-05
Monochrome bulb

gbrooker:
Add the ability to the Lightbulb accessory implementation to represent different types of bulb:

- `.monochrome` : no color variation
- `.color` : can vary hue and saturation
- `.colorTemperature(min, max)` : can vary the color temperature between min and max (reciprocal megakelvin)

and whether the light is on/off only (`isDimmable = false`) or brightness can be varied(`=true`)

Add a precondition to `GenericCharacteristic.init` and `value` setter to verify the new value is within any specified permissible range.

The default values produce a color, dimmable bulb, for compatibility with the current implementation


Bouke:
Looks like an interesting change to have. However I'd like to have a few tests for these, verifying that the bounds work as expected. As the code uses `precondition` to verify the bounds, unit testing those bounds is not possible. So to add unit tests, there probably needs to be some more interface around this. So for now this is fine, I think.

Also, there's some issues with your branch; it shows all prior commits as if you've cherry-picked them. It seems like you merged in my master branch on top of some changes that you've made. This makes it hard for me to see the actual commits that I would need to pull in. I'll probably squash merge this PR, so no need to change things around for now. However I'd appreciate if you could resolve this.

Finally, your git committer name/email is off due to the curly quotes. You might want to verify that you've set them correctly: `git config --global user.name` and `git config --global user.email`.

Tests/HAPTests/EndpointTests.swift
```diff
@@ -552,7 +552,7 @@ class EndpointTests: XCTestCase {
         let lamp = Accessory.Lightbulb(info: .init(name: "Night stand left", serialNumber: "00063"))
 
         lightsensor.lightSensor.currentLight.value = 234
-        thermostat.thermostat.currentTemperature.value = 123
+        thermostat.thermostat.currentTemperature.value = 100 //123 BUGFIX
```
Bouke: Good catch. You can remove the comment.

Sources/HAP/Accessories/Lightbulb.swift
```diff
@@ -37,9 +50,32 @@ extension Service {
             maxValue: 360,
             minValue: 0,
             minStep: 1)
+        public let temperature: GenericCharacteristic<Temperature>?
```
Bouke: The HAP docs also lists brightness, saturation and hue as an optional characteristic. The only required characteristic is on. Can you make those other optional characteristics `Optional<>`?

Sources/HAP/Accessories/Lightbulb.swift
```diff
@@ -37,9 +50,32 @@ extension Service {
...
+            switch type {
+            case .color:
+                characteristics.append(hue)
+                characteristics.append(saturation)
+                temperature = nil
+            case .colorTemperature(let min, let max):
+                precondition(min >= 50 && max <= 400,
+                             "Maximum range for color temperature is 50...400, \(min)...\(max) is out of bounds")
+                temperature = GenericCharacteristic<Temperature>(
+                type: .colorTemperature,
```
Bouke: Please indent these arguments

Sources/HAP/Accessories/Lightbulb.swift
gbrooker: Sure can do, I didn't do that because it could break any code which used Lightbulb as it is, but if you're not so worried about that, I think it is the right change to make. I would also suggest defaulting lightbulb to `.white, isDimmable = false`

gbrooker:
I am having some second thoughts about making these preconditions. Perhaps I should change `precondition` to `assert`, so they will help when debugging, but will not break for a Release build. after the assertions I would then force the value to be within range. Your thoughts ?

gbrooker:
I am still struggling with git and git hub. It seems to be inexplicably complicated on GitHub to have a fork sync with its upstream. I've tried rebasing, but for some obscure reason it doesn't let me on my repository. In the end I have to merge upstream, but this seems to result in this long list of commits. Any hints or pointers would be greatly appreciated.

I'm trying to work out the committer name and email issue too. I didn't have a public email set on GitHub, which might have been the reason. I've fixed that now, but still need to work out how that effects the local git database. Mysteriously Xcode has my GitHub ID three times in the accounts preferences. Where do you see curly quotes ?

Sources/HAP/Accessories/Lightbulb.swift
Bouke: I'm fine with changing behaviour if that means closing the gap with the official spec.

Bouke:
> I am having some second thoughts about making these preconditions. Perhaps I should change precondition to assert, so they will help when debugging, but will not break for a Release build. after the assertions I would then force the value to be within range. Your thoughts ?

I don't think we should force the values to be within range. We should respond with the proper HAP error code notifying that the value was outside of the specified bounds. I'm currently working on a nicer API around Characteristics, which will help in this area as well. For now, we can keep the preconditions until the API has been revised.

gbrooker:
OK, I have made all the optional Characteristics of Lightbulb Optionals, and updated all the tests accordingly. I have made the default bulb `type: .monochrome, isDimmable: false`.

Hopefully my git user name is back to normal. It must have been changed by XCode 9 for some unknown reason, I see it was correct a few weeks ago.

PR #52, opened 2018-03-17
Server.Connection fix race condition

gbrooker:
Writes or socket closure may occur asynchronously on different threads. Race conditions are avoided by performing writes and closure on a private serial queue. Connection().listen() now closes a socket before returning.

### DIFF
--- a/Sources/HAP/Server/Server.Connection.swift
+++ b/Sources/HAP/Server/Server.Connection.swift
@@ -64,6 +64,7 @@ extension Server {
         var cryptographer: Cryptographer?
         var pairing: Pairing?
         var notificationQueue: NotificationQueue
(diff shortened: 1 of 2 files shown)


gbrooker:
There seems to be a problem with the travis build not being able to load your CLibSodium library
From the travis log:
```
...
Fetching https://github.com/Coder-256/Regex.git
error: failed to clone; Cloning into bare repository '/Users/travis/build/Bouke/HAP/.build/repositories/CLibSodium.git--9212624310367625262'...
fatal: unable to access 'https://github.com/Bouke/CLibSodium.git/': Could not resolve host: github.com
...
```

Bouke:
Thanks, looks like a nice change to have in.

PR #56, opened 2018-03-25
Auto-rename HAP Server

gbrooker:
Automatically rename the HAP server if another is running with the same name on the subnet. The Device name property, and the Device's Info.name characteristic will be updated accordingly.

### DIFF
--- a/Sources/HAP/Server/Device.swift
+++ b/Sources/HAP/Server/Device.swift
@@ -7,7 +7,11 @@ fileprivate let logger = getLogger("hap.device")
 
 // swiftlint:disable:next type_body_length
 public class Device {
-    public let name: String
+    internal(set) public var name: String {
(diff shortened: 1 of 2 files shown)


Bouke:
The documentation from Apple is a little sparse on this, but [their example](https://developer.apple.com/library/content/documentation/Networking/Conceptual/NSNetServiceProgGuide/Articles/PublishingServices.html) also notes the same approach:

> Important: As mentioned earlier in this chapter, unless you specifically disable service renaming, your service is automatically renamed if it conflicts with the name of an existing service on the network. If your app depends on the name of the service (whether because its clients store service names for later use or because you display the service name in the user interface), your netServiceDidPublish: method should call the service object’s name method to obtain the name that was actually published.

Thanks for the fix!

PR #61, opened 2018-11-11
Fix for HomePod (add boolean value type)

gbrooker:
Split out from PR #59 

When switching lights on or off, HomePod is sending a boolean value rather than a 1 or 0 which is sent by iOS.

### DIFF
--- a/Sources/HAP/Endpoints/characteristics().swift
+++ b/Sources/HAP/Endpoints/characteristics().swift
@@ -104,9 +104,9 @@ func characteristics(device: Device) -> Application {
 
         case "PUT":
             var body = Data()
-            let byteCount = try? request.readAllData(into: &body)
+            let bytesRead = try? request.readAllData(into: &body)
             logger.debug("PUT data: \(String(bytes: body, encoding: .utf8))")
-            guard byteCount != nil,
+            guard bytesRead != nil,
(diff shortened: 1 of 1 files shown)


Sources/HAP/Endpoints/characteristics().swift
```diff
@@ -104,10 +104,11 @@ func characteristics(device: Device) -> Application {
 
         case "PUT":
             var body = Data()
-            guard
-                (try? request.readAllData(into: &body)) != nil,
-                let decoded = try? JSONDecoder().decode(Protocol.CharacteristicContainer.self, from: body) else
-            {
+            let byteCount = try? request.readAllData(into: &body)
```
Bouke:
```suggestion
            let bytesRead = try? request.readAllData(into: &body)
```

Bouke:
Ah yes, I did read somewhere that "true" and "false" should be valid values as well. However the implementation in iOS did just send 1 and 0 instead, so I probably just forgot about this. I do find it strange that the same platform (Homepod) sends other values.

Sources/HAP/Endpoints/characteristics().swift
gbrooker: Changes applied

PR #63, opened 2018-11-11
Fix unpair, to only change state when no pairings remain

gbrooker:
Split out from PR #59 

When removing  pairing, only change sate when no more pairings remain.

Add unpairFromAllControllers() function to remove all pairings.

### DIFF
--- a/Sources/HAP/Server/Device.swift
+++ b/Sources/HAP/Server/Device.swift
@@ -317,6 +317,22 @@ public class Device {
         return pairingState == .paired
     }
 
+    // Remove all the pairings made with this Device
+    // Can be used in the event of a stale configuration file
+    public func unpairFromAllControllers() {
+        logger.debug("Before unpair")
+        logger.debug(self.configuration)
+        logger.debug(self.config)
+        server?.stop()
+        let allPairingIdentifiers = configuration.pairings.keys
+        for identifier in allPairingIdentifiers {
+            remove(pairingWithIdentifier: identifier)
+        }
+        logger.debug("After unpair")
+        logger.debug(self.configuration)
+        logger.debug(self.config)
+    }
+
     // Add the pairing to the internal DB and notify the change
     // to update the Bonjour broadcast
     func add(pairing: Pairing) {
@@ -339,7 +355,7 @@ public class Device {
             configuration.pairings = [:]
         }
         persistConfig()
-        if pairingState == .paired {
+        if pairingState == .paired && configuration.pairings.isEmpty {
             // swiftlint:disable:next force_try
             try! changePairingState(.notPaired)
         }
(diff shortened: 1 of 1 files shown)


Sources/HAP/Server/Device.swift
```diff
@@ -317,6 +317,22 @@ public class Device {
         return pairingState == .paired
     }
 
+    // Remove all the pairings made with this Device
+    // Can be used in the event of a stale configuration file
+    public func unpairFromAllControllers() {
```
Bouke: We don't notify the other party of the removal, so is it really unpairing or removing the pairing?

```suggestion
    public func removeAllPairings() {
```

Sources/HAP/Server/Device.swift
```diff
@@ -317,6 +317,22 @@ public class Device {
...
+        logger.debug("Before unpair")
+        logger.debug(self.configuration)
+        logger.debug(self.config)
+        server?.stop()
```
Bouke: Why do we stop the server? It's also not clear from the method name that this happens.

Sources/HAP/Server/Device.swift
```diff
@@ -317,6 +317,22 @@ public class Device {
         return pairingState == .paired
     }
 
+    // Remove all the pairings made with this Device
+    // Can be used in the event of a stale configuration file
+    public func unpairFromAllControllers() {
+        logger.debug("Before unpair")
```
Bouke: Do we need all this logging including the before and after state? And if so: can we condense these six lines into a single line?

Sources/HAP/Server/Device.swift
gbrooker: Yes, the HAP protocol doesn't seem to allow for an accessory to request to unpair. The only way to do it is equivalent to a hard reset on a hardware device, i.e to simply close all open connections and forget the pairings. I'll change the function name to your suggestion.

Sources/HAP/Server/Device.swift
gbrooker: Because we need to close all open connections, to force the unpairing.

Sources/HAP/Server/Device.swift
gbrooker: The logging is not needed at all. I can will remove them.

Sources/HAP/Server/Device.swift
Bouke: So what's the exact use-case of this method? I'm guessing it would be some software equivalent of a hardware device's "hard reset". And so, I would expect it to also call `server?.start()`. The documentation would have to be more explicit about what this method does. E.g. should it `start()` when it wasn't already started on entry, or would that be an invalid entry state?

PR #65, opened 2018-11-24
Fix Connection leaks

gbrooker:
As iOS devices move in and out of a network, connections can become stale but HAP never stops listening, even after the same device comes back at a later date and opens a new connection. 

Without this patch in a moderately busy home, HAP will use all available sockets in its process within a couple of days, such that it is no longer able to open any new network connections. Many stale connections to the same devices continue to be held as zombies, with some entering CLOSE_WAIT state after a few days, and the connections never die. Each `Server.Connection` has its own thread on the global queue, so this problem means HAP steadily leaks memory, threads and sockets. 

This pull request addresses these issues with multiple strategies. 

1. Cleanup when closing a connection: whenever a `Server.Connection` is torn down, it removes itself as a listener for any associated characteristics from the bridge `Device`. 

2. Cleanup when removing a pairing: when a pairing is removed, all `Server.Connections` associated with that pairing will be torn down. 

3. Prune dead connections: Every ten minutes, an empty event is sent to tickle each connection (unless there is or have been very recently active communications). This strategy has been borrowed from nfarina's commit to [KhaosT/HAP-Node-JS](https://github.com/KhaosT/HAP-NodeJS/pull/110#issuecomment-132780821) in August 2015.  If the connection is still alive, iOS (the HAP controller) safely ignores the empty event. But if the connection is dead, the Server.Connection is torn down, as either the socket write or the blocking socket read finally detect the remote device has gone or given up.

After this fix, no zombie connections or threads remain after a device leaves the network, or after a pairing is removed. Running HAP with this PR applied for a week in a home with 20 accessories,  3 users and 6 to 8 controllers moving in and out regularly, shows no more thread, memory or socket leaks.

### DIFF
--- a/Sources/HAP/Server/Device.swift
+++ b/Sources/HAP/Server/Device.swift
@@ -361,17 +361,28 @@ public class Device {
     // Remove the pairing in the internal DB and notify the change
     // to update the Bonjour broadcast
     func remove(pairingWithIdentifier identifier: PairingIdentifier) {
-        configuration.pairings[identifier] = nil
-        // If the last remaining admin controller pairing is removed, all
-        // pairings on the accessory must be removed.
-        if configuration.pairings.values.first(where: { $0.role == .admin }) == nil {
-            logger.info("Last remaining admin controller pairing is removed, removing all pairings")
-            configuration.pairings = [:]
-        }
-        persistConfig()
-        if pairingState == .paired && configuration.pairings.isEmpty {
-            // swiftlint:disable:next force_try
-            try! changePairingState(.notPaired)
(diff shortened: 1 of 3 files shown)


Bouke:
Ah yes, this is a very nice addition indeed. Lingering dead connections is something we shouldn't have, and the procedure in this PR looks like a good change.

gbrooker:
I just realised I need to modify the unit tests to deal with this PR. Trying to find a solution right now

Bouke:
Ah yes, I missed that as well. I have setup GitHub to require checks to pass now, it was missing in the repo config.

PR #66, opened 2018-11-24
Update Server.Connection to permit use in unit testing

gbrooker:
Modifications to Server.Connection and MockConnection to allow a test instance to be created without network connections.


Bouke:
What if we make the server argument optional in the constructor of `Server.Connection`?

gbrooker:
For normal usage it should always have a Server argument. It is only when performing the unit tests that we need this fudge.

PR #68, opened 2018-12-15
Rewrite network stack to Swift NIO (armv7 support!)

Bouke:
The current network stack uses BlueSocket, which operates very directly on sockets, which means we have to build our own eventing mechanisms. Swift has seen the release of Swift NIO, an event-driven network application framework. Incorporating NIO means that we outsource a lot of the socket handling and eventing.

This PR incorporates a small portion of Vapor's HTTP handler, to parse the raw HTTP fields received from NIO.

The great thing about this rewrite, is that I'm finally able to run HAP on my Raspberry Pi3 (Stretch) on the community build of Swift 4.1.3.

This PR is not finished, there might be some cases that are not yet handled. I'm currently running this in my home to identify gaps, but so far seems to work like a charm. I plan on doing additional refactors to clean up, but maybe I'll leave that for a future PR.

Todo before merging:
- [x] Verify controller authorization
- [x] Verify pairing authorization role (in `pairings()`)
- [x] Remove pairing + close connection
- [x] Re-enable all sync tests
- [ ] Re-enable all async tests


gbrooker:
Why did HAP not run on the Pi before ? Is BlueSocket incompatible ?

Bouke:
Swift has various 32 bit incompatibilities, and as a result it would segfault a lot; either during compilation or at runtime. There’s a community build version for 4.1.3 that’s much better. Maybe it would even run master branch, I haven’t checked.

Sources/HAP/Server/Server.swift
```diff
@@ -10,148 +11,99 @@ import func Evergreen.getLogger
...
         precondition(device.server == nil, "Device already assigned to other Server instance")
         self.device = device
 
-        application = root(device: device)
+        device.controllerHandler = ControllerHandler()
+        device.controllerHandler!.removeSubscriptions = device.removeSubscriberForAllCharacteristics
+
+        let applicationHandler = ApplicationHandler(responder: root(device: device))
+
+        group = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
```
tgu: I would suggest to only use one thread and thus avoid the need for thread safe code. At least make the number of threads selectable by a function argument. Many use cases will have low traffic and no need for more then one thread.

Bouke: Yes, good idea.

PR #69, opened 2018-12-18
Fix unowned self in Server.addNewConnection

gbrooker:
In some cases the Server instance can be deallocated when the connection closes. Changing unowned self to weak self resolves the issue.


gbrooker:
The travis build check does not seem to be configured for Swift 4.2, which allows `if let self = self`

Bouke:
Swift 4.1 is still a target as that's the latest community build available for Raspbian Stretch. Further while I'm not against this change as such, I'm working towards merging the `swift-nio-http` branch, which replaces `Server` and other connection handling logic.

tgu:
#- ~~~~

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #72, opened 2019-02-03
Initial support for iOS 12.2 (beta) Television Accessory

gbrooker:
Based on KhaosT's HAP-NodeJS implementation, this PR adds the Services and Characteristics for the new Television accessory in the current iOS 12.2 beta.

In order to properly map characteristics, this PR includes support for UInt8, UInt16 and UInt32 based Characteristics. 

Support is added to Service for Linked, Hidden and Primary services.

In order to support the remote control function, I found that the tear down of server connections was too aggressive, so this has been scaled back. I'm running a few instances to test if this reintroduces the port/connection leak that existed before.

### DIFF
--- a/Sources/HAP/Base/Service.swift
+++ b/Sources/HAP/Base/Service.swift
@@ -85,26 +85,43 @@ public extension ServiceType {
     static let garageDoorOpener = ServiceType(0x41)
     static let lockMechanism = ServiceType(0x45)
     static let securitySystem = ServiceType(0x7E)
+    // Television support
+    static let television = ServiceType(0xD8)
+    static let inputSource = ServiceType(0xD9)
+    static let televisionSpeaker = ServiceType(0x113)
 }
 
-open class Service: JSONSerializable {
+open class Service: NSObject, JSONSerializable {
     public let type: ServiceType
 
     weak var accessory: Accessory?
 
     var iid: InstanceID = 0
     let characteristics: [Characteristic]
+    var linkedServices = WeakObjectSet<Service>()
+    var primary: Bool?
+    var hidden: Bool?
 
     public init(type: ServiceType, characteristics: [AnyCharacteristic]) {
         self.type = type
         self.characteristics = characteristics.map { $0.wrapped }
+        super.init()
         self.verify()
     }
 
     init(type: ServiceType, characteristics: [Characteristic]) {
         self.type = type
         self.characteristics = characteristics
+        super.init()
         self.verify()
+    }
+
+    func addLinkedService(_ link: Service) {
+        linkedServices.addObject(object: link)
+    }
+
+    func removeLinkedService(_ link: Service) {
+        _ = linkedServices.remove(link)
     }
 
     func verify() {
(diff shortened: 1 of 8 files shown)


gbrooker:
The travis configuration is failing due to some dependency errors

Sources/HAP/Base/Service.swift
```diff
@@ -85,28 +85,45 @@ public extension ServiceType {
...
     static let lockMechanism = ServiceType(0x45)
     static let securitySystem = ServiceType(0x7E)
+    // Television support
+    static let television = ServiceType(0xD8)
+    static let inputSource = ServiceType(0xD9)
+    static let televisionSpeaker = ServiceType(0x113)
 }
 
-open class Service: JSONSerializable {
+open class Service: NSObject, JSONSerializable {
```
Bouke: Why inherit from NSObject? In order to support Linux, we probably cannot do this.

Bouke:
Looking good thus far, great to see these new features added in! Do you have some television set to test things out?

Sources/HAP/Base/Service.swift
```diff
@@ -85,28 +85,45 @@ public extension ServiceType {
...
     public let type: ServiceType
 
     weak var accessory: Accessory?
 
     var iid: InstanceID = 0
     let characteristics: [Characteristic]
+    var linkedServices = WeakObjectSet<Service>()
+    var primary: Bool?
+    var hidden: Bool?
```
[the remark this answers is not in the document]
gbrooker: These are Optional (linked & primary) or Conditional (hidden) properties of Service, described in sections 2.3.2, 2.3.4 and 5.3.2 of the HAP Spec. 

The Television Service requires the linked property to connect the InputSource Services to the Television. A StatelessProgrammableSwitch also needs to use the linked property too to connect multiple services to one accessory, though I've not experimented with that one. 

Accessories with multiple services are recommended to identify one as a primary service. 

A hidden service is not shown in the Home app.

Sources/HAP/Base/Service.swift
gbrooker: Extending NSObject simply brings in Hashable conformity for free. Hashable conformity is required by WeakObjectSet.

I note that both Server and Server.Connection both already extend NSObject. The first case is in order for the Server to be a NetServiceDelegate. The second is also to use Server.Connections in a WeakObjectSet.

If HAP currently works under linux, then it should be OK to make Service extend NSObject too.

gbrooker:
I have written a back-end which communicates with a Logitech Harmony Hub, a kind of IR blaster configured for various AV setups. The Hub is configured with various 'activity's such as "Watch Netflix" which sets up the TV, set-top boxes, audio etc.

For iOS 12.1 and lower, I simply expose the Hub as a series of switches, one for each 'activity'.
[https://www.brooker.mobi/harmony-bridge/](https://www.brooker.mobi/harmony-bridge/)
(I'm still in a testing phase so only available in a couple of countries app stores for now: FR & SG).

With this PR, I can now expose each Hub as a Television to HomeKit, with each activity as an input source. In iOS 12.2 beta there is also a remote control feature accessed via control centre, which provides remote control keys for up/down/left/right/select/back/play/exit/info. I send these keys to the relevant set top box, via the Harmony Hub. Even the phones volume controls can be used to adjust the TV volume when the remote control view is open on the phone.

I'll be releasing the updated version of the bridge app to the Mac App store soon, though I prefer to wait for at least beta 2 of iOS 12.2 to see if there are any major changes afoot.

Bouke:
That's very cool indeed! I might even have a look at intercepting my Samsung Tv's remote control app, there might be something to it.

PR #78, opened 2019-02-23
Generated from macOS 10.14.4b3

gbrooker:
Update to the Generated HAP structures from 10.14.4 beta 3. This has changes required to support the new Television accessory type supported by iOS 12.2

### DIFF
--- a/Sources/HAP/Base/Generated.swift
+++ b/Sources/HAP/Base/Generated.swift
@@ -319,7 +319,8 @@ public class Enums {
 	public enum CurrentVisibilityState: UInt8, CharacteristicValueType {
 		case shown = 0
 		case hidden = 1
-	}
+        case state2 = 2
+        case state3 = 3	}
 
 	public enum FilterChangeIndication: UInt8, CharacteristicValueType {
 		case noChange = 0
--- a/Sources/HAP/Base/Generated.swift
+++ b/Sources/HAP/Base/Generated.swift
@@ -2,9 +2,9 @@
 // framework definitions. Don't make changes to this file directly.
 // Update this file using the `hap-update` tool.
 //
-// Generated on:              14 February 2019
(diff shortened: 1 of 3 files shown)


gbrooker:
In the latest beta's the current visibility state characteristic has a range of 0 to 3. I've added a couple of characteristics to the Generated.swift and to the Inspector which creates it.

I have also added the correct characteristic settings required to get a Television accessory recognised by iOS 12.2b3.

Sources/HAP/Base/Generated.swift
```diff
@@ -314,6 +316,12 @@ public class Enums {
 		case auto = 1
 	}
 
+	public enum CurrentVisibilityState: UInt8, CharacteristicValueType {
+		case shown = 0
+		case hidden = 1
+        case state2 = 2
```
Bouke: Did you add these by hand?

Bouke:
Overall looks good to me. I guess it's hard to say whether the current TV Accessory configuration is finalized by Apple?

Sources/HAP/Base/Generated.swift
Bouke: The indentation is off and uses spaces (however I would rather use spaces throughout).

gbrooker: I did add these two lines manually to `Generated.swift` for simplicity, but I also added the change to `Inspector.swift` so the same result would be generated in this way automatically next time.

gbrooker:
I think there is at least another beta due soon, as there seem to be a few bugs in the current remote interface. No changes to the HAP definitions between iOS 12.2 beta 2 and beta 3. I'll keep this PR updated until 12.2 final is released, and then we can merge it into the master.

gbrooker:
Replaced by PR #82

Issue #77, opened 2019-02-23
NIO stack - decryption failing


gbrooker:
I'm moving my project over to the new master, but am seeing issues with the NIO stack. I am able to pair with a controller, but HAP gets into an infinite loop reverifying connections over and over, preceded by the following message:

```
[hap.nio|WARNING] Could not decrypt message from [IPv6]::ffff:10.0.8.144/::ffff:10.0.8.144:62546, closing connection.
```

Here is a log extract after pairing. The bridge worked fine in the same HomeKit network and config with the previous network stack. Any pointers on where to look in the NIO code for what is causing this problem ?

```
[hap.controllers.pair-verify|DEBUG] startRequest result: [(HAP.PairTag.identifier, 17 bytes), (HAP.PairTag.signature, 64 bytes)]
[hap.controllers.pair-verify|DEBUG] startRequest result: [(HAP.PairTag.identifier, 17 bytes), (HAP.PairTag.signature, 64 bytes)]
[hap.controllers.pair-verify|DEBUG] startRequest encrypted result: [(HAP.PairTag.state, 1 bytes), (HAP.PairTag.publicKey, 32 bytes), (HAP.PairTag.encryptedData, 101 bytes)]
[hap.controllers.pair-verify|DEBUG] startRequest encrypted result: [(HAP.PairTag.state, 1 bytes), (HAP.PairTag.publicKey, 32 bytes), (HAP.PairTag.encryptedData, 101 bytes)]
[hap.endpoints|INFO] [IPv6]::ffff:10.0.8.8/::ffff:10.0.8.8:50174 POST /pair-verify 200 140
[hap.endpoints|INFO] [IPv6]::ffff:10.0.8.129/::ffff:10.0.8.129:61615 POST /pair-verify 200 140
[hap.controllers.pair-verify|DEBUG] --> username EC48906B-6BEB-4265-97E7-08ADD3E17285
[hap.controllers.pair-verify|DEBUG] --> signature bce6131ba373bc8398c55959d3ca62380975a39a58839d0b1b7023f1b46562e7089039e31d9afd897105d3dbee294715d942fe90455e0f869e2a94b86f8b470d
[hap.controllers.pair-verify|DEBUG] --> public key effeff36f977b12c5ecf7cc536d3bcf9eabc41295768fa88815f4249d205c042
[hap.controllers.pair-verify|INFO] Pair verify completed
[...]

PR #79, opened 2019-02-23
Ensure access to characteristics happen on the main queue. 

gbrooker:
Two small changes. 

First change corrects the precondition on set characteristic value.

Secondly, the characteristic endpoint is currently running on a NIO queue, causing thread safety issues. 

This PR performs the endpoint on the main queue, to ensure a single queue access the HAP characteristics. A future improvement could be to restructure the entire endpoint architecture to use promises, which would avoid a synchronous call across queues.

### DIFF
--- a/Sources/HAP/Base/Characteristic.swift
+++ b/Sources/HAP/Base/Characteristic.swift
@@ -75,7 +75,7 @@ public class GenericCharacteristic<T: CharacteristicValueType>: Characteristic, 
                 return
             }
             precondition(
-                !permissions.contains(.read) || value != nil,
+                !permissions.contains(.read) || newValue != nil,
                 "Readable characteristics should have non nil value")
 
             _value = clip(value: newValue)
--- a/Sources/HAP/Endpoints/root().swift
+++ b/Sources/HAP/Endpoints/root().swift
@@ -26,7 +26,8 @@ func logger(_ application: @escaping Responder) -> Responder {
         let response = application(context, request)
         // swiftlint:disable:next line_length
         logger.info("\(context.channel.remoteAddress?.description ?? "N/A") \(request.method) \(request.urlString) \(response.status.code) \(response.body.count ?? 0)")
-        return response
+        logger.debug("- Response Messagea: \(String(data: response.body.data ?? "nil".data(using: .utf8)!, encoding: .utf8) ?? "-")\n")
+       return response
     }
 }
 
--- a/Sources/HAP/Endpoints/root().swift
+++ b/Sources/HAP/Endpoints/root().swift
@@ -26,7 +26,7 @@ func logger(_ application: @escaping Responder) -> Responder {
         let response = application(context, request)
         // swiftlint:disable:next line_length
         logger.info("\(context.channel.remoteAddress?.description ?? "N/A") \(request.method) \(request.urlString) \(response.status.code) \(response.body.count ?? 0)")
-        logger.debug("- Response Messagea: \(String(data: response.body.data ?? "nil".data(using: .utf8)!, encoding: .utf8) ?? "-")\n")
+        logger.debug("Response Message: \(String(data: response.body.data ?? "nil".data(using: .utf8)!, encoding: .utf8) ?? "-")\n")
        return response
     }
 }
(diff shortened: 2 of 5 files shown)


gbrooker:
Added workaround mentioned in #77 to ensure a minimum buffer size larger than 1155 for messages from HK.

Also added the debug message in root() to log the message sent to HomeKit to help fix issues in communications or configuration. The original version has a spelling mistake ("messagea") which was useful for locating it in the logs, so I preserved the spelling mistake !

gbrooker:
The Tests were running on the main queue, which doesn't;t happen under normal operations. I've modified the characteristic tests to run from a separate queue, simulating a NIO thread.

Tests/HAPTests/EndpointTests.swift
```diff
@@ -143,7 +145,7 @@ class EndpointTests: XCTestCase {
             XCTAssertEqual(nameCharacteristic["value"] as? String, "Night stand left")
         }
 
-        do {
+        testQueue.async {
```
Bouke: 1. How do we ensure correct ordering of multiple items posted on the queue?
2. How do we ensure we actually fail the tests from something happening on another queue?
3. Howe do we ensure the async tests actually finish before finishing the test suite?

Sources/HAP/Endpoints/root().swift
```diff
@@ -26,7 +26,8 @@ func logger(_ application: @escaping Responder) -> Responder {
         let response = application(context, request)
         // swiftlint:disable:next line_length
         logger.info("\(context.channel.remoteAddress?.description ?? "N/A") \(request.method) \(request.urlString) \(response.status.code) \(response.body.count ?? 0)")
-        return response
+        logger.debug("- Response Messagea: \(String(data: response.body.data ?? "nil".data(using: .utf8)!, encoding: .utf8) ?? "-")\n")
```
Bouke: There's a typo and the `- ` prefix is IMO not needed. I think that this would generate non-printable output, especially when the initial pairing is going on. That might result in corrupting the tty session the server is running in, something that we should not do. We could show the http response headers and optionally, depending on the content type, the response's body.

Sources/HAP/Base/Characteristic.swift
```diff
@@ -75,7 +75,7 @@ public class GenericCharacteristic<T: CharacteristicValueType>: Characteristic,
                 return
             }
             precondition(
-                !permissions.contains(.read) || value != nil,
+                !permissions.contains(.read) || newValue != nil,
```
Bouke: oops!

Tests/HAPTests/EndpointTests.swift
gbrooker: 1. The queue is serial, so although each test is run asynchronously, they are executed in sequence. 
2. The call to `application(...)` is run synchronously on the main queue, (from `characteristic(...)`) and returns an `HTTPResponse` back to the test queue where it is examined for correctness. i.e the tests are guaranteed to be performed after the `application(...)` call is complete.
3. We can add the following before the end of each test function
```
        let expectation = self.expectation(description: "Test Complete")
        testQueue.async { expectation.fulfill() }
        waitForExpectations(timeout: 5, handler: nil)
```

BTW I've not dug in to NIO architecture yet, but a better design would be to have NIO call `application()` asynchronously and return a promise to process the result once complete.

Sources/HAP/Endpoints/root().swift
gbrooker: As I mentioned, this is putting back a logging line that has been present in the repository for a long time before the NIO integration, and preserves the original formating, including the spelling mistake from the old version ! I found it useful when searching logs, but we can indeed correct it. 
I would recommend keeping at least some kind of prefix, again for searching debug logs.

The `String(data:, encoding: .utf8)` initialiser returns nil if the unicode is not valid, so we get a `-` line in the log during pairing.

Bouke: > As I mentioned, this is putting back a logging line that has been present in the repository for a long time before the NIO integration, and preserves the original formating, including the spelling mistake from the old version !

I missed that, sorry!

> I found it useful when searching logs, but we can indeed correct it. I would recommend keeping at least some kind of prefix, again for searching debug logs.

Agree.

> The `String(data:, encoding: .utf8)` initialiser returns nil if the unicode is not valid, so we get a `-` line in the log during pairing.

Hmm ok. We could extend this to at least print the headers, but this is also fine I think.

Tests/HAPTests/EndpointTests.swift
Bouke: Ok; looks good to me!

Sources/HAP/Endpoints/root().swift
gbrooker: Fixed the Response log message prefix

PR #80, opened 2019-02-27
Fix/decrypt incomplete frames

Bouke:


gbrooker:
Apologies, I've only just been able to try this out this weekend, and I confirm that this does seem to solve the problem. 

In order to make the test case that I mentioned in Issue #77 work, there are a few unrelated changes required to update the Television accessory for the final release. I will post those in a separate PR.

PR #82, opened 2019-03-30
Update for final release of macOS 10.14.4 and iOS 12.2

gbrooker:
Update for final release of macOS 10.14.4 and iOS 12.2, which include support for the Television accessory.

### DIFF
--- a/Sources/hap-server/main.swift
+++ b/Sources/hap-server/main.swift
@@ -115,7 +115,7 @@ withExtendedLifetime([delegate]) {
         RunLoop.main.run(until: Date(timeIntervalSinceNow: 10))
     } else {
         while keepRunning {
-            RunLoop.current.run(mode: .default, before: Date.distantFuture)
+            RunLoop.current.run()
         }
     }
 }
--- a/Sources/hap-server/main.swift
+++ b/Sources/hap-server/main.swift
@@ -115,11 +115,7 @@ withExtendedLifetime([delegate]) {
         RunLoop.main.run(until: Date(timeIntervalSinceNow: 10))
     } else {
         while keepRunning {
-#if os(macOS)
             RunLoop.current.run(mode: .default, before: Date.distantFuture)
-#elseif os(Linux)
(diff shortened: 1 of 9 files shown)


gbrooker:
This is not compiling because I included a minor change in `hap-server` which is required by swift 4.2. Is it now possible to change the build to swift 4.2 or are there still issues on linux meaning you need to stick to swift 4.1 ?

Bouke:
I'm using Swift 4.2 now on my Raspberry, so I'm fine with dropping 4.1. 👍 It would be great if we could explicitly target 4.2 and 5.0 from now on.

Sources/hap-server/main.swift
```diff
@@ -115,7 +115,7 @@ withExtendedLifetime([delegate]) {
         RunLoop.main.run(until: Date(timeIntervalSinceNow: 10))
     } else {
         while keepRunning {
-            RunLoop.current.run(mode: .default, before: Date.distantFuture)
+            RunLoop.current.run()
```
Bouke: This is not equivalent [see: run() RunLoop](https://developer.apple.com/documentation/foundation/runloop/1412430-run):

> If you want the run loop to terminate, you shouldn't use this method. Instead, use one of the other run methods and also check other arbitrary conditions of your own, in a loop. A simple example would be: (...)

gbrooker: You are right, I've just reverted this change

Bouke:
FYI: I've started a branch to see what's needed to support Swift 5. See also #84.

gbrooker:
So strange, it seems osx and linux swift 4.2 do not have the same definition of RunLoop

Bouke:
Yeah that’s the big downside to Swift Foundation, it’s functionality wise not at par with the OSX counterpart and API wise also lagging behind on changes in the OSX counterpart.

Bouke:
I've merged Swift 5 support, including a fix for the Foundation naming conflict.

gbrooker:
This should now be ready to be merged

PR #84, opened 2019-03-31
Support Swift 5.0

Bouke:


gbrooker:
Great. How about dropping Evergreen.logger and replacing with [swift-log](https://github.com/apple/swift-log)

Bouke:
swift-log would be great, however it target Swift 5, and 4.2 is still the latest and greatest for ARM. So when that’s resolved, I’m all for switching to swift-log.

gbrooker:
Indeed, as part of the move to swift 5. 

I see that NIO2 is released with Swift 5, will that have much impact on your NIO code in HAP ?

Bouke:
I dunno, haven't investigated the differences. NIO2 also targets Swift 5, which prevents me from targetting that version just yet.

PR #88, opened 2019-04-24
Fix thread safety in pairing endpoints and ChannelHandlers

gbrooker:
I've been seeing hard crashes due to unsafe access to shared dictionaries from multiple NIO threads in these functions. I've added code to ensure all accesses are thread-safe. The sessions array may be read concurrently, but writes are exclusive.

Regarding the TODO comment on not freeing memory you wrote, should we setup a timeout to clear sessions if the pairing does not complete, or a background process to flush out old sessions once per hour ?

### DIFF
--- a/Sources/HAP/Endpoints/pairSetup().swift
+++ b/Sources/HAP/Endpoints/pairSetup().swift
@@ -30,8 +30,23 @@ func pairSetup(device: Device) -> Responder {
                                           algorithm: algorithm))
     }
 
-    // TODO: this memory is not freed, not thread-safe either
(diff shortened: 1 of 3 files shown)


Bouke:
Thanks for the fix!

> Regarding the TODO comment on not freeing memory you wrote, should we setup a timeout to clear sessions if the pairing does not complete, or a background process to flush out old sessions once per hour ?

Any setup is fine I guess; where the timeout would probably be the overall lower on resources as there won't be unneeded background cleanup going on.

Issue #49, opened 2018-02-04
Maintain instance ids assignments between restarts


Bouke:
From the HAP specification 2.6.1.2:

> After a firmware update services and characteristics types that remain unchanged must retain their previous instance ids, newly added services and characteristics must not reuse instance ids from services and characteristics that were removed in the firmware update.

We currently reassign instance ids.

gbrooker:
This should be fixed with the addition of the persistent configuration

Bouke:
So currently, changing the ordering of services or acessories, would change the instance identifiers. As we already have some persistence in place, I think this can be fixed by extending on that.

gbrooker:
This should be closed now

PR #89, opened 2019-08-18
Persist Serial numbers

gbrooker:
When removing accessories from a Bridge, the current implementation also removes the serial number to AID mapping. If the accessory is added back, a new AID is generated, which is not in line with the HAP spec.

This PR adds an optional parameter `andForgetSerialNumbers:` to `Device.removeAccessories()` to persist the serial number to AID mapping after removing accessories. 

A default value is provided to ensure any existing code operates as before. However I would suggest it may be better to make persistence the default, and only remove the serial number mapping if specifically requested to do so. (i.e making the default value true). What do you think ?

A host application can permanently forget all serial number to AID mappings by removing the Storage.

### DIFF
--- a/Sources/HAP/Server/Device.swift
+++ b/Sources/HAP/Server/Device.swift
@@ -275,7 +275,7 @@ public class Device {
         server?.updateDiscoveryRecord()
     }
 
-    public func removeAccessories(_ unwantedAccessories: [Accessory]) {
+    public func removeAccessories(_ unwantedAccessories: [Accessory], andForgetSerialNumbers: Bool = true) {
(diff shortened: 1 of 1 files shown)


Bouke:
This would also fix #49 I think?

gbrooker:
OK, I changed the default behaviour to not remove the mapping of serial number to AID.

It seems that the linux configuration in the Travis builds has an issue with dependancies.

Bouke:
I’m not sure what’s up with the Linux build. Looks like some dependency is broken. I’ll have a look some other time, as it is unrelated to this PR. Thanks!

Issue #100, opened 2020-02-09
TargetHumidifierDehumidifierState wrong enum


makleso6:
Hi!
I think TargetHumidifierDehumidifierState is wrong enum
```
public enum TargetHumidifierDehumidifierState: UInt8, CharacteristicValueType {
    case inactive = 0
    case idle = 1
    case humidifying = 2
}


public enum CurrentHumidifierDehumidifierState: UInt8, CharacteristicValueType {
    case auto = 0
    case humidifyAuto = 1
    case dehumidifyAuto = 2
}
```

in HomeKit iOS native implementation enum represents as

```
public enum HMCharacteristicValueTargetHumidifierDehumidifierState : Int {

    case automatic

    case humidify

    case dehumidify
}
public enum HMCharacteristicValueCurrentHumidifierDehumidifierState : Int {

    case inactive

    case idle

    case humidifying

    case dehumidifying
}
```

Bouke:
Our enums are being generated from the Homekit framework that's bundled with macOS. I've just verified, but it seems that our enums are correct based on that file. However, to be sure, I also checked the parts from the HAP specifications:

> 9.116 Target Humidifier Dehumidifier State
>
> Property|Value
> -|-
> UUID|000000B4-0000-1000-8000-0026BB765291
> Type|public.hap.characteristic.humidifier-dehumidifier.state.target
> Permissions|Paired Read, Notify, Paired Write
> Format|uint8
> Minimum Value|0
> Maximum Value|2
> Step Value|1
> Valid Values|0 ”Humidifier or Dehumidifier”<br>1 ”Humidifier”<br>2 ”Dehumidifier”

And

> 9.29 Current Humidifier Dehumidifier State
>
> Property|Value
> -|-
> UUID|000000B3-0000-1000-8000-0026BB765291
> Type|public.hap.characteristic.humidifier-dehumidifier.state.current
> Permissions|Paired Read, Notify
> Format|uint8
> Minimum Value|0
> Maximum Value|3
> Step Value|1
> Valid Values|0 ”Inactive”<br>1 ”Idle”<br>2 ”Humidifying”<br>3 ”Dehumidifying”

These are indeed in line with what you've written. After some investigation, the reason was found:

```xml
<key>HUMIDIFIER_DEHUMIDIFIER_MODE</key>
<dict>
    <key>Format</key>
    <string>string</string>
    <key>OutValues</key>
    <dict>
        <key>0</key>
        <string>INACTIVE</string>
[...]

PR #101, opened 2020-02-09
fix CurrentHumidifierDehumidifierState and TargetHumidifierDehumidifierState

makleso6:


makleso6:
issue #100

Bouke:
Thank you for your investigation and PR. We don't accept manual changes to the generated file, as those will be lost next time the file is generated. I've pushed a fix in 88066c2573fdd03f3c371064f85fc3e767e135f2 that resolves the problem with auto generation of this file.

PR #104, opened 2020-05-04
EventLoopGroup can be set by Server.init

Michael-mhr:
In case someone is managing its own EventLoopGroup as I do in my project RoomCtrlClient4

### DIFF
--- a/Sources/HAP/Server/Server.swift
+++ b/Sources/HAP/Server/Server.swift
@@ -29,7 +29,7 @@ public class Server: NSObject, NetServiceDelegate {
     public init(
         device: Device,
         listenPort requestedPort: Int = 0,
-        numberOfThreads: Int = System.coreCount
+        worker: EventLoopGroup? = nil
     ) throws {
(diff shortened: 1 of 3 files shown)


Sources/HAP/Server/Server.swift
```diff
@@ -38,8 +38,13 @@ public class Server: NSObject, NetServiceDelegate {
...
 
         let applicationHandler = ApplicationHandler(responder: root(device: device))
-
-        group = MultiThreadedEventLoopGroup(numberOfThreads: numberOfThreads)
+        
+        if let worker = worker {
+            self.group = worker
+        } else {
+            self.group = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
+        }
```
Bouke, line 46: I haven't written Swift for some time, but shouldn't this also work?
```suggestion
        self.group = worker ?? MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
```

Bouke:
Thanks, merged in 6f1a29eb40e34cb95b863b4301bb281ac939f632!

PR #105, opened 2020-05-16
Allow Int based Characteristic values to be initialised by Float or Double

gbrooker:
An interesting bug came up recently, when initialising a `colourTemperature` characteristic, where the default value in the factory method in `Generated.swift` was outside the max and min range specified in my init call. The value of the characteristic is an `Int` type, but the min and max values in `Generated` are stored as `Double`. Creating the characteristic with that new range resulted in a crash, as the double max an min values could not be converted to `Int`

This patch adds the ability to create an `Int` characteristic from a `Double` or `Float`.


Bouke:
Interesting bug indeed! [Travis build seems fine][1], despite the status not showing up here.

[1]: https://travis-ci.org/github/Bouke/HAP/builds/687698528

PR #107, opened 2020-05-25
add valid values and valid values range for Characteristic

makleso6:


Sources/HAP/Base/Characteristic.swift
```diff
@@ -50,6 +52,11 @@ extension Characteristic {
         if let maxValue = maxValue { serialized["maxValue"] = maxValue }
         if let minValue = minValue { serialized["minValue"] = minValue }
         if let minStep = minStep { serialized["minStep"] = minStep }
+        if let validValuesRange = validValuesRange {
```
makleso6, line 55: Should use range otherwise array?

makleso6: @Bouke

Bouke: The specification notes regarding ranges the following:

> This descriptor shall be used when it is more efficient to specify a list of valid values as a range rather than individual values.

So, this looks fine to me.

PR #121, opened 2020-12-03
Structure channel handler additions in a more readable way

lorenzofiamingo:


Bouke:
What qualifies readable depends on the individual I guess. Your suggested code has less nesting, which is good. However due to varying uses of `addHandlers`, `flatMap` and nested `addHandlers`, I'm not sure the resulting pipeline is the same. The correctness isn't communicated by the code. I'm therefore inclined to close this PR.

PR #123, opened 2020-12-26
Update Generated to macOS 11.1

gbrooker:
Updated Generated to the latest macOS 11.1 version. Added a default value for UInt16 types.

### DIFF
--- a/Sources/HAP/Base/Generated.swift
+++ b/Sources/HAP/Base/Generated.swift
@@ -2,9 +2,9 @@
 // framework definitions. Don't make changes to this file directly.
 // Update this file using the `hap-update` tool.
 //
-// Generated on:              25 may 2020
-// HomeKit framework version: 838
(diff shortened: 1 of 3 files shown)


Bouke:
I haven't upgraded to Big Sur (yet), how's your experience thus far? I think I might wait for 11.2 before biting the bullet.

gbrooker:
> I haven't upgraded to Big Sur (yet), how's your experience thus far? I think I might wait for 11.2 before biting the bullet.

I'm running into enough issues that I would recommend holding off if you can. WiFi was dropping regularly, which was fixed by resetting the network location. Now I am getting random pauses where the machine becomes unresponsive for a few seconds, and sometimes the window manager seems to block. I'm running on a 2019 16" MacBookPro, so it is not as though the machine is out of date.

Bouke:
Good to know, will definitely be holding off for now!

PR #126, opened 2021-01-04
Make AccessoryType, ServiceType and CharacteristicType's CustomStringConvertible

gbrooker:
Added a description computed property to make AccessoryType, ServiceType and CharacteristicType objects into CustomStringConvertible, and add a logging bootstrap to hap-server main.swift


Sources/HAP/Base/Generated.swift
```diff
@@ -44,6 +44,47 @@ public enum AccessoryType: String, Codable {
...
+			"29": "Faucet",
+			"30": "Shower Head",
+			"31": "Television",
+			"32": "Target Controller",
+			"33": "Wi-Fi Router",
+			"34": "Audio Receiver",
+			"35": "Television Set Top Box",
+			"36": "Television Streaming Stick",
+        ]
+        return descriptions[self.rawValue] ?? ""
```
RLovelett, line 86: Since these are generated files, rather than silently failing would it be better to `fatalError`?

The reason I ask is this seems like it would signal that there was an unexpected type and likely needs to be regenerated.

Sources/HAP/Base/Generated.swift
```diff
@@ -96,6 +137,61 @@ public extension ServiceType {
 	static let windowCovering = ServiceType(0x008C)
 }
 
+extension ServiceType: CustomStringConvertible {
+    public var description: String {
```
Bouke, line 141: I would prefer a switch-statement here instead with cases similar to the definition, like so:
```swift
    public var description: String {
        switch self {
        case let .appleDefined(hex):
            switch hex {
            case 0x0239: return "Accessory Runtime Information"
            case 0x00BB: return "Air Purifier"
            // ...
            default: fatalError("Unknown Apple-defined service type '\(hex)'")
            }
        case let .custom(uuid):
            return "Other \(uuid)"
        }
    }
```

Sources/HAP/Base/Generated.swift
```diff
@@ -44,6 +44,47 @@ public enum AccessoryType: String, Codable {
 	case audioReceiver = "34"
 	case televisionSetTopBox = "35"
 	case televisionStreamingStick = "36"
+
+    public var description: String {
```
Bouke, line 48: I prefer to have a switch-statement here instead.

Sources/HAP/Base/Generated.swift
```diff
@@ -2,13 +2,13 @@
...
 //
-// Generated on:              26 December 2020
+// Generated on:              3 January 2021
 // HomeKit framework version: 858
 // macOS:                     Version 11.1 (Build 20C69)
 
 import Foundation
 
-public enum AccessoryType: String, Codable {
+public enum AccessoryType: String, Codable, CustomStringConvertible {
```
Bouke, line 11: Can you move the `CustomStringConvertible` to an extension, similar to what you've done with `ServiceType`?

Sources/HAP/Base/Generated.swift
```diff
@@ -254,6 +350,215 @@ public extension CharacteristicType {
 	static let wiFiConfigurationControl = CharacteristicType(0x022D)
 }
 
+extension CharacteristicType: CustomStringConvertible {
```
Bouke, line 353: Same comment as `ServiceType.description`.

Sources/hap-server/main.swift
```diff
@@ -11,6 +11,26 @@ fileprivate let logger = Logger(label: "bridge")
     import Glibc
 #endif
 
+func createLogHandler(label: String) -> LogHandler {
```
Bouke, line 14: I guess this changes the log level depending on the release? Can you move that to a separate PR?

Bouke (changes requested):
Overall a good change and will improve diagnostics.

Sources/HAP/Base/Generated.swift
gbrooker: Most certainly not. This file defines currently known and non-blacklisted services and characteristics only. It is highly likely others will be used and/or defined in the future that this package doesn't know about, at least yet.

The swifty way would be to return an optional, but this is a public standard API so I chose to return an empty string for unknown properties, which can be equally well tested for by the caller.

Sources/HAP/Base/Generated.swift
gbrooker: OK I have added the top level switch, but a switch and fatalError is not appropriate for the descriptions themselves. It is quite possible that HAP does not have the full knowledge of all valid values.

Sources/HAP/Base/Generated.swift
gbrooker: OK done

Sources/HAP/Base/Generated.swift
gbrooker: OK Same reply

Sources/hap-server/main.swift
gbrooker: OK moved to #127

Sources/HAP/Base/Generated.swift
Bouke: There's only a `fatalError` for the `appleDefined` case. But I agree that you could put a value in there that we're unaware of. So in that case something like `"Apple-defined \(hex)"` would be fine.

gbrooker: Sure, that's in my commit of yesterday

Sources/HAP/Base/Generated.swift
gbrooker: OK given this type is restricted to only specific cases I have made the description use a switch statement

Sources/HAP/Base/Generated.swift
Bouke: I still think it should use a switch, like my suggested code. Instead of a fatalError you could return the value in the default case.

Bouke (changes requested):

Sources/HAP/Base/Generated.swift
Bouke: I still think it should use a switch, like my suggested code. Instead of a fatalError you could return the value in the default case.

gbrooker: While I much prefer the neatness of the dictionary lookup, I have made the change you request for both ServiceType and CharacteristicType.

Sources/HAPUpdate/Inspector.swift
```diff
@@ -366,7 +366,22 @@ public class Inspector {
...
-        write("}")
+        write("""
+        }
+
+        extension AccessoryType: CustomStringConvertible {
+            public var description: String {
+                switch self {
+        """)
+        for category in categoryInfo.sorted(by: { $0.id < $1.id }) {
+            write("\t\tcase .\(categoryName(category.name)):")
```
Bouke, line 377: We use 4 spaces for indentation, not a tab.

Bouke (changes requested):

Sources/HAP/Base/Generated.swift
Bouke: Ah I think there was some filter active as I didn't see those changes.

Sources/HAPUpdate/Inspector.swift
gbrooker: What ? The Inspector has written tab's since its inception, there are 168 occurrences of '\t' in the class....

Sources/HAP/Base/Generated.swift
gbrooker: It's definitely there in commit 64bd0f2cb12418813ddcc81b72edf0f6beb62dba

for both ServiceType and CharacteristicType

Sources/HAP/Base/Generated.swift
```diff
@@ -96,6 +173,115 @@ public extension ServiceType {
 	static let windowCovering = ServiceType(0x008C)
 }
 
+extension ServiceType: CustomStringConvertible {
+    public var description: String {
+        switch self {
+        case let .appleDefined(typeCode):
+            let hex = String(typeCode, radix: 16).uppercased()
```
Bouke, line 180: I'd prefer not to convert the value here, and it seems we can easily prevent this. See my original suggestion that just uses the hex values, similar to how we're defining the cases. 
```swift
    public var description: String {
        switch self {
        case let .appleDefined(hex):
            switch hex {
            case 0x0239: return "Accessory Runtime Information"
            case 0x00BB: return "Air Purifier"
            // ...
            default: fatalError("Unknown Apple-defined service type '\(hex)'")
            }
        case let .custom(uuid):
            return "Other \(uuid)"
        }
    }
```

Sources/HAP/Base/Generated.swift
Bouke: Same remark regarding the hex conversion.

Sources/HAPUpdate/Inspector.swift
Bouke: I see: the file _is_ using tabs. However this PR introduces mixed-style indentation.

This is also a major reason I use space indents everywhere. While I'm fine with this PR using tabs (throughout), I'll probably update the code to replace tabs by spaces to align with the overall code in this project.

Bouke (changes requested):

Sources/HAP/Base/Generated.swift
gbrooker: No problem, just updated

Sources/HAPUpdate/Inspector.swift
gbrooker: OK I've replaced all the \t with four spaces

Bouke (approved):

Bouke (approved):

PR #127, opened 2021-01-05
Add logging bootstrap to select debug messages

gbrooker:


Sources/hap-server/main.swift
```diff
@@ -11,6 +11,26 @@ fileprivate let logger = Logger(label: "bridge")
...
+        default:
+            handler.logLevel = .info
+        }
+    #else
+        // No logging
+        handler.logLevel = .critical
+    #endif
+    return handler
+}
+LoggingSystem.bootstrap(createLogHandler)
```
Bouke, line 32: What's the goal of adding this to the example server? I'd prefer to have most of the logging available, so that potential users of the package can easily evaluate the project. With this change a release build wouldn't print most logs?

gbrooker: The goal is to show someone using the package how to selectively enable and disable logging. It certainly wasn't obvious to me once we switched from the Evergreen logger, which I found much easier to use.

The default settings here will turn >on< debug logging for everything in hap except the verbose encryption logs (which is only to serve as an example for how to selectively disable something).

Bouke: Evergreen was easier to use, I agree. However it was unsupported and this is the package Apple is migrating to.

I'm not a big fan of adding this code to the example code, as it makes the overall file harder to read. Maybe the code could live in some util file, so we only have to call `setupLogging()` from `main.swift`. What do you think?

gbrooker: Sure, I've lifted `createLogHandler()` out to a separate file. I suggest we keep the call `LoggingSystem.bootstrap()` in the main file, as it should only be called as the program starts up.

PR #131, opened 2021-03-16
Only shutdown the EventLoopGroup if it was created by this instance

gbrooker:
Currently, calling `stop()` on a HAP `Server` will shutdown the NIO event loop group, no matter whether it was created by that Server instance or provided by a caller. This PR will ensure that if the EventLoopGroup is provided in the initialiser, and hence most likely shared with other Servers and NIO channels, then stopping the server will not shutdown the event loop.


Bouke (approved):

Bouke:
Nice fix!

PR #153, opened 2022-11-23
New HomeKit architecture compatibility in iOS 16.2

gbrooker:
For compatibility with the new HomeKit architecture introduced in iOS 16.2, Pair Verify step M4 "Finish Response" must correctly adhere to the spec and return an authentication error for certain issues. The new architecture on a HomeHub (HomePod or AppleTV) uses this error to determine which version of keys the accessory uses.

The current response is to return an HTTP error 400, however this cause HomeKit under iOS 16.2 to mark the accessory as unaccessible.

### DIFF
--- a/Sources/HAP/Endpoints/pairVerify().swift
+++ b/Sources/HAP/Endpoints/pairVerify().swift
@@ -63,7 +63,45 @@ func pairVerify(device: Device) -> Responder {
         } catch {
             logger.warning("Could not verify pairing: \(error)")
             setSession(for: context, to: nil)
-            return .badRequest
+
+            // HAP Specification 4.8.4:
+            // Must return authentication error to Verify Finish Response (M4) when
+            // 1. fail verifiction of authentication tag on encrypted keys
+            // 2. decryption fails
+            // 3. pairing ID is not recognised
+            // 4. fail verification of iOS device signature
+
+            let response: PairTagTLV8?
+
+            switch error {
+            case
+                PairVerifyController.Error.couldNotDecrypt,
+                PairVerifyController.Error.couldNotDecode,
+                PairVerifyController.Error.noPublicKeyForUser,
+                PairVerifyController.Error.invalidSignature:
+
+                response = [
+                    (.state, Data([PairVerifyStep.finishResponse.rawValue])),
+                    (.error, Data([PairError.authenticationFailed.rawValue]))
+                ]
+            default:
+                // For other errors, respond with an HTTP error 400 (bad request)
+                // PairVerifyController.Error
+                // .invalidParameters
+                // .noSession
+                // .couldNotSetupSession
+                // .couldNotEncrypt
+                // Ed25519.Error
+                // .invalidSignature
+                // .couldNotSign
+                response = nil
+            }
+
+            if let response = response {
+                return HTTPResponse(tags: response)
+            } else {
+                return .badRequest
+            }
         }
     })
 }


Sources/HAP/Endpoints/pairVerify().swift
```diff
@@ -63,7 +63,45 @@ func pairVerify(device: Device) -> Responder {
         } catch {
             logger.warning("Could not verify pairing: \(error)")
             setSession(for: context, to: nil)
-            return .badRequest
+
+            // HAP Specification 4.8.4:
+            // Must return authentication error to Verify Finish Response (M4) when
```
Bouke, line 68: The documentation you've quoted says that the authentication error should be returned for "Verify Finish Response (M4)", but the implementation will now also do that for the start request. Which is broader than the M4 step. I suppose this is not a problem?

Sources/HAP/Endpoints/pairVerify().swift
```diff
@@ -63,7 +63,45 @@ func pairVerify(device: Device) -> Responder {
...
+            // 1. fail verifiction of authentication tag on encrypted keys
+            // 2. decryption fails
+            // 3. pairing ID is not recognised
+            // 4. fail verification of iOS device signature
+
+            let response: PairTagTLV8?
+
+            switch error {
```
Bouke, line 76: Should we only return an authentication error for this limited set of errors, or could we do it always? If we could, it would simplify things here.

Bouke (commented):
Overall looks good to me and good to go if you can clear up my questions.

Bouke:
> The current response is to return an HTTP error 400, however this cause HomeKit under iOS 16.2 to mark the accessory as unaccessible.

"Pair Verify" is performed to establish a connection. If there's an authentication error, what would happen with this change merged? And how's that different from iOS prior 16.2?

Sources/HAP/Endpoints/pairVerify().swift
gbrooker: This is not an issue, none of the errors caught are thrown in startRequest, so there is no change to behavior during earlier steps. Only the following errors are caught and result in the authentication error response.
                PairVerifyController.Error.couldNotDecrypt,
                PairVerifyController.Error.couldNotDecode,
                PairVerifyController.Error.noPublicKeyForUser,
                PairVerifyController.Error.invalidSignature:

Sources/HAP/Endpoints/pairVerify().swift
gbrooker: To correctly implement the spec requirements, this error should only be raised for these specific errors.

gbrooker:
> > The current response is to return an HTTP error 400, however this cause HomeKit under iOS 16.2 to mark the accessory as unaccessible.
> 
> "Pair Verify" is performed to establish a connection. If there's an authentication error, what would happen with this change merged? And how's that different from iOS prior 16.2?

The change does not impact earlier versions of iOS/tvOS/homepodOS. Either the current behavior of reporting an HTTP error, or the revised behavior of reporting an authentication error (in line with spec) resulted in a failed connection and Home reported that the device was unresponsive.

Under iOS16.2, the Home app proposes to users to upgrade to a home to a new HomeKit architecture including Matter devices with the following message:
```
Upgrade All Homes
Complete the steps to upgrade your homes and take advantage of improved 
performance and reliability of the accessories in your homes.
```
Under the new HomeKit implementation, there are now two sets of pairing keys, the ones used for the legacy HomeKit architecture (HH1) and the new architecture (HH2). When the user migrates, iOS rolls the keys to new HH2 keys and attempt to tear down pairings using the old HH1 keys but it will still fall back to them if an accessory is not responding for whatever reason. 

A HomeHub will attempt to verify using the one set of keys, and if that fails (as the keys have been rolled over), it will retry using the other keys. This is not working under iOS 16.2 (and related tvOS/HomePod s/w)  because it expects HAP to reply with a TLV8 error code on the first attempt so the Home Hub knows to try the other key instead. Under the current implementation, HAP doesn't send any payload in response to the pair-very so the HomeHub never falls back to the other keys. Without the change, HAP devices are marked as unresponsive under Home after Upgrading. With the implementation returning a TLV8 error message when HAP rejects pair-verify, the issue is resolved.

Bouke:
Thank you for the in-depth reply. Looks like a good change to incorporate. 👍

PR #161, opened 2023-01-06
Extra enums

TheMisfit68:
added two extensions for enums (that get used by 'irrigation systems')


Bouke:
Where are these definitions coming from? Do you have an authoritative reference for them?

TheMisfit68:
I found them in Apple's document "HAP-Specification-Non-Commercial-Version.pdf" and tried to implement them following the same patterns as those I found in your framework already.

PR #162, opened 2023-09-21
Added public label to the HAP.Service

TheMisfit68:
Helps to distinguish services of the same type that are part of the same accessory


Bouke:
Can you describe the use-case for this change?

TheMisfit68:
I have several custom Accessories that often have more than one service of a particular type, for instance, I have an accessory for my electric car that has two Switches associated with it, one to start charging and one to start the airconditioning. When I handle events in the delegate I needed to be able to differentiate between the two. 
It's also really handy for debugging purposes.

Issue #163, opened 2023-09-28
Cut a new Release?


mattcorey:
The latest release - 0.7.0 - is from January 2020 (Pre-Pandemic!), and there have been several dozen commits to the 'master' branch since then.  The release version may not even run in Xcode 15 - I have a project that fails to run, because the SRP package claims I have an incompatible OS version.

Is the current 'master' branch active and in use?  Then does it make sense to cut a 0.8.0 release, or even a 1.0.0 release here, to avoid these issues?  After all, the nearly-four-year-old release is the default that will be picked up by Xcode if you don't change to point to a branch

PR #164, opened 2023-11-25
Upped macOS version requirements to v11 to allow compilation on macOS

Diggory:
SRP requires 10_15 and HKDF is only available in 11

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,7 @@ import PackageDescription
 let package = Package(
     name: "HAP",
     platforms: [
-        .macOS(.v10_15),
+        .macOS(.v11),
     ],
     products: [
         .library(name: "HAP", targets: ["HAP"]),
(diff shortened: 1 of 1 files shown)


Diggory:
Hi @Bouke, How do I get the SwiftLink check to complete?  Thanks.

Bouke:
Did you remove your changes from your master branch? There are no changed files in this PR. The SwiftLint check should run automatically, let's see what happens when you add commits.

What are you trying to change? `Package.swift` already has a v11 requirement.

Diggory:
Huh, that's weird - I seem to have changed the V11 from your master to V10_15 then back to V11. 
No idea why I did that!

Diggory:
Ahh, I see what happened -  I tried to depend on HAP using the most recently tagged release.  That failed because it didn't have any platforms minimum requirements.  So I assumed that main was the same, but main is not the same.

Any chance of tagging a new release so we don't have to depend on a moving target?  #163 

p.s. thanks for all your work on HAP - it's very useful.  So useful that I [Nominated it for the Package Community Showcase](https://forums.swift.org/t/nominations-for-the-packages-community-showcase-on-swift-org/68168/4).

